PR #7, opened 2017-10-12
week 1 Bikes-for-refugees webpage recreation without using bootstrap - Prabesh

prabshah:
Hello guys, I just tried to create the exercise 1 page without using bootstrap.


styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+
+*/
+
+/* temp styling */
+
+
+
+/* main styling for code your futrue week 1 tutorial by PRABESH recreation of the page without using bootstrap*/
+header {
+    max-width:85%;
```
matt-paul: It's normal to leave a single space between css properties and the value, eg 'max-width: 85%' rather than 'max-width:85%'

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+    
+    margin-top:20px;
+    padding:20px;
+    border-radius:5px;
+    border-width:0px;
+    background:#CE5F31;
+    color:white;
+    
+    font-size:.9em;
+
```
matt-paul: why the empty lines? 😀

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+    border-radius:5px;
+    border-width:0px;
+    background:#CE5F31;
+    color:white;
+    
+    font-size:.9em;
+      
+}
+
+.thankyou-message-display-box {
```
matt-paul: This is a very long class name, a one or two word name would be easier to remember and be more readable.  You might also think whether there is a more generic name that you could use..

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+    color:#CE5F31;
+}
+.blue-color-learn-more {
+    color:#0275D8;
+}
+.footer-section {
+    margin-top: 50px;
+}
+.footer-section > p {
+    font-size:1.04em;
```
matt-paul: Incredibly precise font-size. 😂 Did you mean this?

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+.events-pics {
+    height:100px;
+    width:100px;
+}
+.event-description {
+    margin-left:20px;
+}
+.event-headings {
+    font-weight:bold;
+    font-size:.9em;
```
matt-paul: Favour 0.9em over.9em as it more readable

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+    padding-top:-20px;
+    margin-top:-45px;
+}
+.horizontal-line-inside-bike-pic {
+ opacity:.5;
+ 
+}
+.buttons-inside-pic {
+    margin-top:15px;
+    padding:15px 15px;
```
matt-paul: Look up the shorthand version of this (line 126)

index.html
```diff
@@ -1,3 +1,114 @@
...
+           <li><a href="">NEWS & EVENTS</a></li>
+           <li><a href="">CONTACT US</a></li>
+       </ul>
+       <button class="donate-now flex-item">Donate Now</button>
+   </header>
+   <div class="thankyou-message-display-box">
+       <p>You've helped us give 72 bikes to refugees and asylum seekers. <strong>Thank you!</strong></p>
+   </div>
+   <section class="main-body-content">
+       <div class="left-section">
```
matt-paul: How would you make this html more semantic?  A number of places where this could be improved...

matt-paul:
Extra note:  When you fix these, often best to commit these things one issue at a time.  Eg, you might go through you're whole project, fix all spaces between css properties and values, and then add the relevant css files and commit with a clear 'Add space between css properties and values' commit message.  Then move onto the next issue, and commit again.  This makes it very easy for people to see what you have done.  Also, try to start each commit with a verb in the present tense...'Add...', 'Fix...', 'Remove...' etc

styles/style.css
```diff
@@ -1,10 +1,217 @@
...
+    width:158px;
+}
+.navigation-ul {
+    list-style:none; 
+}
+.navigation-ul li {
+    float:left;
+    width:10%;
+}
+.navigation-ul li:nth-child(3), .navigation-ul li:nth-child(4) {
```
matt-paul: This is difficult to read...how about adding a class to the two navigation elements you are targeting?

index.html
```diff
@@ -1,3 +1,113 @@
...
+            <br><br>
+            <strong>LEARN MORE</strong>
+            <hr class="lines line-two">
+            <div class="articles-left">
+                <div class="article-one">
+                    <p><strong>Why do refugee need bikes?</strong></p>
+                    <p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo
+                         ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis    
+                    </p>
+                    <span class="orange-triangle">&#9658;</span><span class="learn-more">Learn more</span>
```
OlenaKashuba: How did you make the triangle?

index.html
```diff
@@ -1,3 +1,113 @@
...
+            <div class="footer-section">
+                <hr class="lines line-three">
+                <p>
+                    Bikes for Refugees Scotland is a volunteer-led project to help people explore and make connection with their local comminities.  
+                </p> 
+            </div>   
+       </div>
+       <div class="right-section">
+          <P>UPCOMING EVENTS</P>
+          <hr class="lines line-four">
```
OlenaKashuba: you can create a line using pseudo elements + css, but hr tag is fine as well

styles/style.css
```diff
@@ -1,10 +1,230 @@
...
+.navigation-ul {
+    list-style: none; 
+}
+
+.navigation-ul li {
+    float: left;
+    width: 10%;
+}
+
+.navigation-ul li:nth-child(3), .navigation-ul li:nth-child(4) {
```
OlenaKashuba: why do set different width for li-elements?

PR #16, opened 2018-04-09
using grid system

than2042:
Please merge!

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -31,11 +31,33 @@ body {
 	font-weight: 600;
 	text-transform: uppercase;
 }
+.nav-link {
+margin: 10px;
+padding: 5px 5px;
+}
+.nav-link {
+	border: 1px solid transparent;
+
+  }
+.nav-link:hover,
+.nav-link:focus {
+	background-color: red;
+}
+
+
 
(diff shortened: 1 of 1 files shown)


styles/style.css
```diff
@@ -31,11 +31,37 @@ body {
...
+.btn-secondary {
+	background-color: rgb(228, 235, 164);
+	border-radius: 7px;
+}
+
+.btn-primary {
+background: #ce5f31;
+border: none;
+border-radius: 10px;
+}
```
LucyMac: the btn in the top right corner has too much height.

index.html
```diff
@@ -61,34 +61,47 @@ <h1 class="display-3">Bikes for Refugees</h1>
...
-								</a>
-							</p>
-						</article>
-					</div>
-				</div>
+						</div>
+						
 				<div class="sidebar-content col-lg">
 					<h2 class="heading-underline">Upcoming events</h2>
 					<div class="articles">
```
LucyMac: the 3 items under Upcoming Events are not aligned with the heading.
Hint: Avoid using `row` without `col-` classes inside - it does strange things!

PR #20, opened 2018-04-11
Newchanges

Seraphineyoung:
Work in progress

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -37,7 +37,7 @@ body {
 	border-radius: 0.15em;
 }
 
-/* Seraphine code starts here, I just wanted to create a branch and merge with master,made more changes */
+/* Seraphine code starts here, I just wanted to create a branch and merge with master. Made more changes*/
 .btn-primary {
 	background-color: #ce5f31;
 }
--- a/styles/style.css
+++ b/styles/style.css
@@ -37,7 +37,7 @@ body {
 	border-radius: 0.15em;
 }
 
-/* Seraphine code starts here, I just wanted to create a branch and merge with master */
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -65,8 +93,41 @@ body {
...
 .jumbotron {
 	border-radius: 0;
+	/*background-image: url(../images/header-bike.jpg);*/
+    background-size: cover;
+    background-repeat: no-repeat;
+	background-position: center bottom;
+	color: white;
+	background: linear-gradient( rgba(0, 0, 0, 0.7), rgba(0, 0, 0, 0.7) ), url('../images/header-bike.jpg');	
+	opacity: 0.8;
+}
```
LucyMac: very nice effect on the image!

styles/style.css
```diff
@@ -65,8 +93,41 @@ body {
...
+	background-color: lightgray;
+	padding: 2rem;
+
+}
+
+article.bg-light {
+	border: 1px solid #939393;
+	padding: 2rem;
+	margin-left: 2rem;
+}
```
LucyMac: You should be adding horizontal padding to the same level as your `col-...` classes as it will break the grid. The trick to avoid the 2 columns touching, is to move all content, borders and background colours to one level down (or move the `col-` up one level). You'll need to add a wrapping `div` somewhere to do this...

styles/style.css
Seraphineyoung: Thanks Lucy for the review..

styles/style.css
Seraphineyoung: Okay..thanks for the feedback will look into this now

PR #21, opened 2018-04-11
Francisca V

franciscavargas:


index.html
```diff
@@ -3,24 +3,24 @@
...
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css" integrity="sha384-rwoIResjU2yc3z8GV/NPeZWAv56rSmLldC3R/AZzGRnGxQQKnKkoFVhFQhNUwEyJ" crossorigin="anonymous">
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
 		<link rel="stylesheet" href="styles/style.css">
 	</head>
 	<body>
 		<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
-		<div class="site-header container">
-			<div class="navbar navbar-toggleable-md navbar-light">
+		<header class="site-header container">
+			<nav class="navbar navbar-toggleable-md navbar-light">
```
LucyMac: the navbar will need a `row` class to be aligned to the rest

styles/style.css
```diff
@@ -124,4 +142,46 @@ body {
...
+
+.bike-more-info {
+	padding: 6rem 0;
+}
+
+.bike-more-info .card {
+	min-height: 250px;
+	justify-content: center;
+	align-items: center;
+}
```
LucyMac: Nice. You could add a margin-bottom on mobile so they are not stuck together once they stack.

index.html
```diff
@@ -61,84 +61,116 @@ <h1 class="display-3">Bikes for Refugees</h1>
...
+								<p class="article-read-more">
+									<a href="#">
+										<span class="fa fa-caret-right"></span>
+										Learn more
+									</a>
+								</p>
+							</div>
+						</article>
+
+						<article class="article card col-sm-12 col-lg-5">
```
LucyMac: You can't have horizontal padding/margins on custom classes where you also use `col-...` classes, it undoes what the grid does! 🙂 
To avoid the boxes touching, move the border/background color one level down (or move the `col-` class one level up. You will have to add a `div` or similar...

styles/style.css
```diff
@@ -37,6 +37,24 @@ body {
 	border-radius: 0.15em;
 }
 
+.btn-primary {
+		background-color: #ce5f31;
+}
```
LucyMac: Watch out when using Bootstrap's ready-made buttons - they come with default styling such as the blue border, which you may want to overwrite here.

PR #24, opened 2018-04-12
End of Homework

KarwanM:


index.html
```diff
@@ -52,18 +52,30 @@ <h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and ac
 			</div>
 			<div class="row">
 				<div class="primary-content col-lg-8">
-					<div class="jumbotron">
+					<div class="jumbotron col-lg-12">
+						<div class="back-contents col-lg-12">
```
LucyMac: Be careful here, you are nesting `col-...` classes directly one inside the other. You cannot do this.

The order must be:

1) `container` (grandparent). 
You can use this just once to wrap around your whole webpage if you want, and then only have rows inside. 
2) `row` (parent). To group blocks together.
3) `col-...` (children) **note:** must add up to 12

Each level must be the direct child of the other.

index.html
```diff
@@ -52,18 +52,30 @@ <h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and ac
...
 						<p class="lead">Providing donated bikes and accessories to refugees and asylum seekers in Scotland.</p>
 						<hr class="my-4">
 						<div class="buttons">
 							<a class="btn btn-primary btn-lrg" href="#">Donate a bike today</a>
 							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
 						</div>
+						</div>
+					</div>
+
+					<div class="three-boxes col-lg-12">
```
LucyMac: You cannot add any custom horizontal padding, margin or width to the same level as a `col-` class as it will interfere with it.

index.html
```diff
@@ -52,18 +52,30 @@ <h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and ac
...
+						<button class="col-lg-3">Check Availability</button>					
+						<button class="col-lg-3">Donate Bikes</button>
```
LucyMac: Again, you are nesting column classes directly. They need to be separated by `row`:

```
<div class="three-boxes col-lg-12">
	<div class="row">
		<button class="col-lg-3">Check Availability</button>					
		<button class="col-lg-3">Donate Bikes</button>
		<button class="col-lg-3">Volunteer with us</button>
	</div>
</div>
```

PR #25, opened 2018-04-13
WIP

Behnazz:
I haven't started the bootstrap bit yet.

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,17 +14,17 @@
 	</head>
 	<body>
 		<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
-		<div class="site-header container">
+		<header class="site-header container">
 			<div class="navbar navbar-toggleable-md navbar-light">
 				<button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
 					<span class="navbar-toggler-icon"></span>
 				</button>
 				<a href="/" class="navbar-brand">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -61,70 +64,121 @@ <h1 class="display-3">Bikes for Refugees</h1>
...
+							
+							<article class="article">
+								<div class=articles-row>
+										<a href="#">
+											<img class="img-fluid" src="./images/spring-fundraisers_thumbnail.jpg">
+										</a>
+							  			<div>
+
+											<div class=push>
+										<div class="article-title article-title--sidebar">
```
LucyMac: the idea here is not to use custom horizontal margins like you do in your `push` class. 
Try and do everything just using the 3 grid classes `container`, `row`,  and `col-...`
You can move these around, add wrappers, remove them - as you wish, but try not to use any custom spacing and sizing styles (horizontal only, you can add top and bottom spacing, and heights).

styles/style.css
```diff
@@ -62,22 +62,84 @@ body {
...
 	font-size: 1rem;
 	font-weight: 600;
 	text-transform: uppercase;
 }
 
+.article-container{
+display: flex;
+flex-wrap: wrap;
+flex-direction: row;
+justify-content: space-between;
```
LucyMac: Instead of laying the 2 boxes out with `space-between` here, try laying them out using `row` and `col-` classes directly on the elements. You will need to think of a trick to avoid the boxes touching 🙂

index.html
```diff
@@ -14,17 +14,17 @@
 	</head>
 	<body>
 		<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
-		<div class="site-header container">
+		<header class="site-header container">
 			<div class="navbar navbar-toggleable-md navbar-light">
```
LucyMac: Your navbar will beed a `row` class here as it is directly inside a container.

PR #26, opened 2018-04-13
three boxes add

brook-code:


styles/style.css
```diff
@@ -66,12 +66,31 @@ body {
 .jumbotron {
 	border-radius: 0;
 }
+container {
+	display: flex;
+	width: 100%;
+	height: auto;
+	flex-direction: row;
+
```
LucyMac: this class is missing a `.` at the beginning - but aside from this, be careful: `.container` is already a Bootstrap class so overwriting it can cause unexpected behaviour.

index.html
```diff
@@ -61,6 +61,14 @@ <h1 class="display-3">Bikes for Refugees</h1>
 							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
 						</div>
 					</div>
+					<div class="container row justify-content-between">
```
LucyMac: You cannot use a container and a row on the same level. The order must be: 
1) `container` (grandparent)
2) `row` (parent)
3) `col-...` (children) **note: must add up to 12**
each level must be the direct child of the other.

LucyMac: If you follow this, structure, your 3 boxes will align with the rest of that block.

PR #29, opened 2018-04-14
bisrat

Bisratta:


index.html
```diff
@@ -103,20 +118,20 @@ <h2 class="heading-underline">Upcoming events</h2>
...
 						<article class="article row">
 							<a href="#">
-								<img class="img-fluid" src="https://raw.githubusercontent.com/Code-Your-Future/syllabus/master/html-css/assets/bikes-for-refugees-logo.jpg">
+								<img class="img-fluid" src="images/bikes-for-refugees_logo.jpg" style=" width:50px; height:50px;">
 							</a>
 							<div class="article-title article-title--sidebar">
 								<a href="#">Bike dropoff event in Edinburgh on May 11</a>
 							</div>
 						</article>
 						<article class="article row">
```
LucyMac: Watch out, you cannot have a `row` without `col-...` classes inside (on the children elements)

index.html
```diff
@@ -13,17 +13,19 @@
 		<link rel="stylesheet" href="styles/style.css">
 	</head>
 	<body>
-		<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
-		<div class="site-header container">
+      <div class="container my-container">
+          <div class="row my-row">
+              <div class="col-md-10 my-col">
```
LucyMac: `my-row` and `my-col` are not needed here.

index.html
```diff
@@ -39,28 +41,41 @@ <h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and ac
...
 						<div class="buttons">
-							<a class="btn btn-primary btn-lrg" href="#">Donate a bike today</a>
-							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
+							<a class="btn btn-lrg blue-button" href="#">Donate a bike today</a>
+							<a class="btn volunteer btn-lrg" href="#">Volunteer</a>
 						</div>
-					</div>
+          </div>
+          <div class="parentBox">
+            <div class="box"> check availability </div>
```
LucyMac: These look good. Just be careful with your choice of names. Be more descriptive than `box`, for example `feature` or `highlight` or `action-box`. And for parent elements, if they have no other use than wrapping around, you can call them `...-wrapper`, for example `feature-wrapper`

PR #30, opened 2018-04-14
Boxes created

Tebesi:
I have created three boxes in a row.

### DIFF
--- a/index.html
+++ b/index.html
@@ -110,7 +110,7 @@
 					<div class="articles">
 						<article class="article row">
 							<a href="#">
-								<img class="img-fluid" src="home/coder/desktop/images/spring-fundraisers_thumbnail.jpg">
+								<img class="img-fluid" src="http://res.cloudinary.com/dsskzinju/image/upload/v1523702915/spring-fundraisers_thumbnail.jpg">
 							</a>
 							<div class="article-title article-title--sidebar">
 								<a href="#">Join us for our Spring fundraisers</a>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -61,6 +61,22 @@ <h1 class="display-3">Bikes for Refugees</h1>
 							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
 						</div>
 					</div>
+					<div class="container my-container ">
+						<div class="row my-row justify-content-between">
+							<div class="box col-md-3 col-12 my-col "><strong>
```
LucyMac: `my-col` and `my-row` are not used here, please remove them. 
As an aside, you should not be using classes which a) overwrite the Bootstrap grid classes (`container`, `row` and `col-...`), as this would interfere with the grid's layout; and b) which are named `my-something`, it's not descriptive and when you work on shared code with other people it loses its meaning.

index.html
```diff
@@ -94,24 +110,24 @@ <h2 class="heading-underline">Upcoming events</h2>
 					<div class="articles">
 						<article class="article row">
```
LucyMac: If you use a `row` class, the elements inside it *must* have `col-...` classes. The row is there to work with columns, if you use it by itself it will misalign the layout.

index.html
LucyMac: Same thing with `my-container`. Be more descriptive.

styles/style.css
```diff
@@ -78,6 +78,28 @@ body {
 	text-transform: uppercase;
 }
 
+.my-container {
+	width: 100%;
+	height:auto;
```
LucyMac: You don't need this because you're already using Bootstrap's `container` class. All you need is to add the `padding-bottom`, and you should do that in a more descriptive class name.

PR #32, opened 2018-04-14
Submit for review 2

MHDGhanemBalhawan:
I'm trying to solve my git problem. Hope you could compare the changes now.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,6 +2,7 @@
 <html lang="en_US">
 	<head>
 		<title>Bikes for Refugees | Scotland</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1">
 		<meta charset="utf-8">
 		<link rel="dns-prefetch" href="//fonts.googleapis.com">
 		<link rel="dns-prefetch" href="//maxcdn.bootstrapcdn.com">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -61,9 +61,17 @@ <h1 class="display-3">Bikes for Refugees</h1>
 							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
 						</div>
 					</div>
+					<div class="container row justify-content-around">
```
LucyMac: You cannot use a container and a row on the same level. The order must be: 
1) `container` (grandparent)
2) `row` (parent)
3) `col-...` (children) **note: must add up to 12**

styles/style.css
```diff
@@ -88,33 +103,44 @@ body {
 	max-width: 100px;
 	background: #ce5f31;
 }
-
+.container {
+	flex: 1 1 30%;
```
LucyMac: You cannot write your own `container` class when using Bootstrap. It will interfere with the grid.

styles/style.css
```diff
@@ -88,33 +103,44 @@ body {
...
+}
+.box {
+  height: 150px;
+  width: 150px;
```
LucyMac: ⚠️ You use the `box` class on the same element as a `col-md-3` class... 

https://github.com/CodeYourFuture/bikes-for-refugees/pull/32/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR67

which one will win?

styles/style.css
```diff
@@ -88,33 +103,44 @@ body {
...
+  text-align: center;
+  margin: 1% auto 3% auto;
+  padding: 1%;
+}
 /* Article */
 .article {
 	margin-bottom: 2em;
+	border: 1px solid gray;
+	background-color: #D8FFD6;
+	margin:1% 1%;
```
LucyMac: Same thing, you use the `article` class on the same level as `col-md-5`: `article` defines its own margin which will cause problems with the grid. Also, you are putting the border/background-color directly on the `col-` element, which will render unexpectedly. Bootstrap is for invisible boxes - if you want the box to be visible, that needs to happen on a child element inside the column.

index.html
```diff
@@ -94,7 +102,7 @@ <h2 class="heading-underline">Upcoming events</h2>
 					<div class="articles row">
 						<article class="article">
```
LucyMac: You cannot have a `row` without `col-...` classes on the child elements.

LucyMac:
Good work opening the PR by the way!! 😃

index.html
MHDGhanemBalhawan: I'll do it again.
Thank you for your review.

PR #33, opened 2018-04-14
Than bikes

than2042:
Hi Lucy, 
Please have a look! thank you.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,6 +2,7 @@
 <html lang="en_US">
 	<head>
 		<title>Bikes for Refugees | Scotland</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta charset="utf-8">
 		<link rel="dns-prefetch" href="//fonts.googleapis.com">
 		<link rel="dns-prefetch" href="//maxcdn.bootstrapcdn.com">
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -90,10 +119,33 @@ body {
...
+	font-weight: 700;
+	text-align: center;
+	margin: 1rem;
+}
+
 .article {
 	margin-bottom: 2em;
 }
 
+.articles .element {
```
LucyMac: please use a more descriptive name than 'element'

index.html
```diff
@@ -61,9 +62,16 @@ <h1 class="display-3">Bikes for Refugees</h1>
 							<a class="btn btn-secondary btn-lrg" href="#">Volunteer</a>
 						</div>
 					</div>
+					<div class="container">
```
LucyMac: Be careful using containers. They are not really meant to be nested like this. Use one around you're whole webpage, and then just work with `row`s and `col-...`s

styles/style.css
```diff
@@ -90,10 +119,33 @@ body {
...
+	
+}	
+
+.my-box .box {
+	border: 1px solid #292b2c;
+	min-height: 7rem;
+	font-size: 27px;
+	font-weight: 700;
+	text-align: center;
+	margin: 1rem;
```
LucyMac: You cannot add margin on a level that also has a `col-` class, it will interfere with the grid.

index.html
```diff
@@ -94,24 +102,23 @@ <h2 class="heading-underline">Upcoming events</h2>
 					<div class="articles row">
```
LucyMac: if you use a `row` you **must** have `col-...` elements inside. 

Just like you do here: https://github.com/CodeYourFuture/bikes-for-refugees/pull/33/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR74

index.html
```diff
@@ -75,7 +83,7 @@ <h2 class="heading-underline">Learn more</h2>
 								</a>
 							</p>
 						</article>
-						<article class="article">
+						<article class="article element col-md-5 col-sm-12">
```
LucyMac: You cannot use column classes and add custom widths as well:

https://github.com/CodeYourFuture/bikes-for-refugees/pull/33/files#diff-3e1a93b11aefa9b36204aaec251b6741R145

PR #40, opened 2019-03-16
Clean up and instructions

LucyMac:
**Markup clean up and instructions**

There was quite a bit of 'bad practice' in the code which would contradict what we're teaching, and a lot of Bootstrap clutter, so I've cleaned up the markup a bit.

Also introduced some BEM notation.

I added clear instructions for the exercise, including an additional, non-visual task to replace `div`s with semantic markup where possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Then, **clone** it to your machine, using the following terminal commands.
 `cd <your-project-directory>`
 
 >Clone the repo:
-`git clone git@github.com:<your_username>/bikes-for-refugees.git`
+`git clone https://github.com/<your_username>/bikes-for-refugees.git`
 
 Now, **create a new branch** with your name in the title
 >`git checkout -b bikes-for-refugees/yourname`
@@ -52,5 +52,6 @@ _Make sure you're on your branch_
 
 > In your terminal: add, commit and push your changes
 
-> Go to Github and open a PR
+> Go to Github and open a PR **with the CodeYourFuture repository as the base repository** (not your own).
 
+Look here for help with your PR: https://codeyourfuture.github.io/syllabus-master/others/making-a-pull-request.html
--- a/README.md
+++ b/README.md
@@ -1,12 +1,56 @@
 # bikes-for-refugees
-Exercise for HTML lesson 1
+Exercise for HTML & CSS - lesson 1
 
-Fork the repository to your personal Github account and use the following terminal commands to download the files to your projects folder.
+### Set up
 
-`cd ~`
+**Fork** this repository to your personal Github account (look for the 'fork' button on this page).
+
+Then, **clone** it to your machine, using the following terminal commands.
+
+>Go to the folder you wish to save the project in:
 `cd <your-project-directory>`
+
+>Clone the repo:
 `git clone git@github.com:<your_username>/bikes-for-refugees.git`
 
-**Note:**
-Some of the links to images are broken. The correct images have been added in the `images` folder - please replace any broken image links with paths to the images that are in the `images` folder.
+Now, **create a new branch** with your name in the title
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,12 +1,56 @@
...
-`cd ~`
+**Fork** this repository to your personal Github account (look for the 'fork' button on this page).
+
+Then, **clone** it to your machine, using the following terminal commands.
+
+>Go to the folder you wish to save the project in:
 `cd <your-project-directory>`
+
+>Clone the repo:
 `git clone git@github.com:<your_username>/bikes-for-refugees.git`
```
lpalma: Not all students have SSH keys configured and we had few problems in the last module because of that.
It might be better to clone with `https` here

40thieves: We also could link to https://codeyourfuture.github.io/syllabus-master/others/making-a-pull-request.html too?

README.md
```diff
@@ -1,12 +1,56 @@
...
+### Exercise
+
+The aim of the exercise is for your webpage to look like this: 
+https://codeyourfuture.github.io/syllabus-london/html-css/assets/screenshot-complete.png
+
+Open the `index.html` file in your browser, and compare it to the image of the final result. What's different?
+
+#### Use semantic markup
+
+Where necessary, replace the `<div>`s with semantic HTML tags. This will not render differently in the browser, but it's good practice to use semantic markup when you can!
```
lpalma: will the students already be familiar with _semantic html_ when they start the exercise?

README.md
```diff
@@ -1,12 +1,56 @@
...
-Some of the links to images are broken. The correct images have been added in the `images` folder - please replace any broken image links with paths to the images that are in the `images` folder.
+Now, **create a new branch** with your name in the title
+>`git checkout -b bikes-for-refugees/yourname`
+
+You're now ready to code!
+
+### Exercise
+
+The aim of the exercise is for your webpage to look like this: 
+https://codeyourfuture.github.io/syllabus-london/html-css/assets/screenshot-complete.png
```
40thieves: Probably out of scope for this PR, but I wonder if we could deploy a completed version somewhere. Perhaps github pages might be good for that?

README.md
LucyMac: Good point! Thanks

README.md
LucyMac: Yes 😄

README.md
```diff
@@ -1,12 +1,56 @@
...
+#### Fix the positioning
+
+Use Flexbox to move elements around so they are positioned correctly.
+
+https://codeyourfuture.github.io/syllabus-london/html-css/assets/screenshot-complete.png
+
+
+### Submit your work
+
+Once you've finished, **open a pull request** so your work can be reviewed by a mentor.
```
lpalma: It may be worth to specify where to open the PR, i.e. to their own fork or to the original project.

README.md
LucyMac: Yeah but we don't want them inspecting the code now do we ;)

40thieves: Yeah I guess that's a tradeoff. Github pages would hide it under the `gh-pages` branch I believe

README.md
```diff
@@ -1,12 +1,57 @@
...
+
+_Make sure you're on your branch_
+
+> In your terminal: add, commit and push your changes
+
+> Go to Github and open a PR **with the CodeYourFuture repository as the base repository** (not your own).
```
LucyMac: @lpalma I've added this line but don't feel I can add more here without it getting confusing 😬  
I'll try and go over this in class...

lpalma: yeah that's a good point. I can't think of any good way to put that in words. They've done it in the past module anyways, so that should be fine

PR #46, opened 2019-03-20
Kadir-Bikes-for-Refugees-project

laperla11:


styles/style.css
```diff
@@ -91,10 +116,53 @@ button {
...
+	height:100px;
+	width:auto;
+	display:block;
+}
+
+.sub-content {
+	display : flex;
+	justify-content:space-between;
+}
+.sub-content .article {
```
mattyates52: `border-color` should be `grey` or try changing the border `opacity`

laperla11:
> Nice work Kadir - I would say the `font-sizes` throughout the page are a little too big. I have made a few other comments, but other than that, it looks great!

Hi Matt, your feedback is very much appreciated. I made the required correction and pushed it.

PR #49, opened 2019-03-23
Elamin Bike-For-Refugee

aldeez988:


styles/style.css
```diff
@@ -6,136 +6,180 @@
...
-	   max-height: 80px;
-	}
-}
\ No newline at end of file
+ html,
+ body {
+   font-family: "Source Sans Pro", "Helvetica Neue", Arial, sans-serif;
+ }
+ 
+ button {
```
mattyates52: Trying removing the `border` on your buttons, and adding some `border-radius` to match the design. Have a look at removing the `outline` and changing the `cursor` when hovering a button for extra points!

styles/style.css
```diff
@@ -6,136 +6,180 @@
...
+   max-height: 40px;
+   width: auto;
+ }
+ 
+ .navbar-light .navbar-toggler {
+   border: 0;
+   right: 0;
+ }
+ 
+ .navbar-light .navbar-nav .nav-link {
```
mattyates52: I think the `font-size` could be reduced slightly, and some `margin-right` could be added for a bit of extra spacing

styles/style.css
```diff
@@ -6,136 +6,180 @@
...
+   position: absolute;
+   left: -1px;
+   top: 0;
+   height: 100%;
+   width: 1px;
+   background: #ce5f31;
+ }
+ 
+ /* Jumbotron */
+ .jumbotron {
```
mattyates52: I think you should increase the height of the `jumbotron`. You can use `height` or increasing the `padding` may look better

aldeez988: Thank you @mattyates52  for your review and you feedback

PR #50, opened 2019-03-24
sola-bike-project

solasobowale:


index.html
```diff
@@ -90,38 +91,45 @@ <h6 class="article__title">
 									</a>
 								</p>
 							</article>
+						</section>
 
 					</div>
 				</div>
 
 				<div class="sidebar col-lg-4">
```
goduffy22: The children in this sidebar need a bit better positioning. Add `display: flex; flex; flex-flow: flex-wrap;` to wrap it's immediate children THEN (see below how to style children):

index.html
```diff
@@ -90,38 +91,45 @@ <h6 class="article__title">
...
 						<h2 class="heading-underline">Upcoming events</h2>
-						<article class="article">
+						<section class="fundraisers1">
 							<a href="#">
```
goduffy22: Add a class to all the `<a>` tags that contain an image to size the images inside better. Use `width: 30%' and then on the text next to it do 'width: 70%' so that all the images line up at 30% across each section and all the text boxes line up too.

index.html
```diff
@@ -90,38 +91,45 @@ <h6 class="article__title">
...
-								<img class="img-fluid" src="https://raw.githubusercontent.com/Code-Your-Future/syllabus/master/html-css/assets/news-spring-fundraisers.jpg" alt="">
-					  	</a>
+								<img class="img-fluid" src="./images/spring-fundraisers_thumbnail.jpg" alt="spring-logo">
+						  </a>
+						<article class="article3">
```
goduffy22: Add `width: 70%` to this class and the same for the other articles

index.html
```diff
@@ -43,7 +43,7 @@ <h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and ac
 							<a class="nav-link" href="#">Contact Us</a>
 						</li>
 					</ul>
-					<button>Donate Now</button>
+					<button class = "donate-button">Donate Now</button>
```
goduffy22: You probably haven't covered this yet but there are button classes in bootstrap which will style this button really nicely like in the example with rounded corners etc. Try adding the class `btn` to all the buttons if you want to see how it styles it. Here's a link with some of the other possible classes https://www.w3schools.com/bootstrap4/bootstrap_buttons.asp

goduffy22:
Well done Sola. The styling was really good, only a few small things to fix. Good job.

PR #73, opened 2021-05-15
North West 4-Anza Azam-HTML/CSS-Week1

Anza-Azam:
bike-for-refugees_fixed-webpage

### DIFF
--- a/index.html
+++ b/index.html
@@ -81,7 +81,7 @@
       <div class="row">
         <div class="main col-lg-8">
           <div class="alert">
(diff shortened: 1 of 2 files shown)


Anza-Azam:
Completed with flex layout after Week 2.

KarenPudner (approved):
Good use of semantic HTML. The links all work and the buttons look correct. The sections resize and change position when the page is resize, which shows a good use of flex. Just 2 small points - there should be a small gap between the images and text on the right hand sections, and you should correct any errors in the text - there should be a gap between "give" and "72". But a good piece of work.

PR #74, opened 2021-05-15
Revise base code

Morteza-Khojasteh:
Revise index.html & Style.css

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,7 +100,7 @@
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
+            <div class="article-content">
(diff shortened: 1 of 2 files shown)


lannem (commented):
Good work Morteza!

To improve you could use some more semantic tags in your HTML. For example some of the div tags could be section tags to make them more descriptive

Morteza-Khojasteh:
> Good work Morteza!
> 
> To improve you could use some more semantic tags in your HTML. For example some of the div tags could be section tags to make them more descriptive

Thanks, Lucy. I'll do that

PR #75, opened 2021-05-16
Myaskylar-bikes-for-refugees

myaskylar:


KarenPudner (approved):
Good use of semantic HTML. The CSS targets classes rather than IDs which is good. The elements of the page move and resize effectively when the page is resized, showing a good use of flex. The buttons look correct and the navigation links go orange on hover as requested. The link to the main image is broken - your css file is in a subfolder so the image path needs to take that into account.  And you should check the text for any errors - there should be a space between "give" and "72'. Also, I am not sure why you have created the grey boxes - is this part of the instructions that I have missed? But overall, a good piece of work.

myaskylar:
Thanks for the review Karen. I've rectified the errors. The grey boxes just a exercise we had in previous class. I've removed it.

PR #79, opened 2021-05-23
NW - Terry Calderbank - Bikes for Refugees - Week 2 HTML/CSS

Texx45:
Finished project ready for marking

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # bikes-for-refugees
+
 Exercise for HTML & CSS - lesson 1
 
 ### Set up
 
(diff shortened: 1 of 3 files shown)


SteveLeicester:
Thanks for completing the bikes for refugees Terry.  The image is fixed and the flexbox is in place along with button stylings. Did you miss completing the semantic stuff?  So I can see <div class="article"> which looks like it should be <article>.   Missing borders around the upcoming events cards too.

Texx45:
Hi Steve,

I think I messed this one up a bit. I was in your group with Alexandra and we worked our way through he first couple but for some reason I did not save it, so when I came to do it I had to start it again.  It looks like I have rushed through it to be honest as I was it was the week we set quite a bit of homework, I can go back and fix it when I have completed this week's homework if that is ok?

Regards,

Terry

PR #80, opened 2021-05-23
fix bikes-for-refugees

Abed-Aghbar:
fix bikes-for-refugees

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,12 +47,7 @@
             src="images/bikes-for-refugees_logo.jpg"
             alt="Logo for Bikes for Refugees Scotland"
           />
-          <h1 class="sr-only">
-            Bikes for Refugees Scotland - Providing donated bikes and
-            accessories to refugees and asylum seekers.
-          </h1>
         </a>
-
         <div
           class="navigation__content collapse navbar-collapse"
           id="navbarSupportedContent"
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -17,6 +17,11 @@ button {
 
 /* Navigation */
 
+.donate_now {
```
mike-shields-dev, line 20: Don't forget to add the common styles to all buttons further down where it says /*INSERT BUTTON STYLES HERE*/

styles/style.css
```diff
@@ -98,20 +110,27 @@ button {
 /* Article */
 
 .article {
-  margin-bottom: 2em;
+  margin-bottom: 1em;
```
mike-shields-dev, line 113: Articles also have a grey border with rounded corners.

index.html
```diff
@@ -73,7 +68,7 @@ <h1 class="sr-only">
               <a class="nav-link" href="#">Contact Us</a>
             </li>
           </ul>
-          <button>Donate Now</button>
+          <button class="donate_now">Donate Now</button>
```
mike-shields-dev, line 71: Hi Abed, you need to go through all of the HTML and where neccesary replace the `<div>` tags with semantic tags such as `<header>`, `<footer>`, `<section>`, `<article>` etc

index.html
```diff
@@ -47,12 +47,7 @@
             src="images/bikes-for-refugees_logo.jpg"
             alt="Logo for Bikes for Refugees Scotland"
           />
```
mike-shields-dev, line 49: Hi Abed, I don't think you should have removed the `<h1>` here. It was there for accesibility of visually impaired people. Although it could not be seen in the browser, it would have been read by a screen reader.

PR #87, opened 2021-05-27
NW4 - Azan Shaikh - Bikes For Refugees - Week 2 - HTML/CSS

Ash-97:


KarenPudner (approved):
Good use of semantic HTML, and you target classes rather than IDs in your css, which is good practice and has allowed you to reuse classes. The buttons look correct. The sections move effectively when the window is resized which shows a good use of flex - however, when the sections on the right hand side move to the bottom of the page when the window is narrowed the middle section is narrower than the other 2 sections. Also you need to check your paths to the main image and the spring challenge image - they are currently broken. And you should check the text you have been given and correct if necessary - there should be a gap between "give" and  "72". But overall, a good piece of work,

PR #93, opened 2021-06-01
NW4 WEEK2 -Bike for refugees- Bimbola-Lasisi

blasisi:
fixed some of the error, Readjust the button section, change navigation colour and adjust article section using css styling.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,230 +1,189 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
 
-    <link rel="dns-prefetch" href="//fonts.googleapis.com" />
-    <link rel="dns-prefetch" href="//maxcdn.bootstrapcdn.com" />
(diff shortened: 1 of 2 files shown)


nax2uk (commented):
Hi @blasisi,
Good Effort on this homework. I can see you have spent some time on it. Please have a look at my feedback and let me know if you have any questions.


**Use semantic markup**
Where necessary, replace the <div>s with semantic HTML tags. This will not render differently in the browser, but it's good practice to use semantic markup when you can!

>> Well done you used the `<header>`. You could use other semantic tags like `<main>`, `<footer>`, `<nav>`, `<article>`

**Fix the broken images**
Some of the links to images are broken. The images you need are in the images folder - please replace any broken image links with the correct paths.

>> Great Job

**Style buttons**
There are 3 buttons on the page. They need to be styled correctly - remember, re-use styles as much as possible by using classes.
>> Good job on the buttons. Perhaps it could be made better by having a hover.
Also, you shouldn't have to have separate class names for different buttons if they look the same. E.g. the `donate-bike-button` class and `donate-button` class looks the same.


**Fix the positioning**
Use Flexbox to move elements around so they are positioned correctly.
>> The donate button should be on the right.
>>  The 'why do refugees need bikes' and 'how can I help' article should have a border around them and both should have a white background.
>> The 'Learn More' links should be formatted the same.

PR #99, opened 2021-06-04
Exercise for HTML & CSS - lesson 1

mmahareeq:


SteveLeicester:
Thanks for competing the bikes for refugees Maziar.  You have the images in place and are saving space with the flexbox.  Styling of buttons is as expected - good.  Care with your semantic HTML.  I think I saw <artical class=   but it should be <article   > An easy spelling mistake and I think most browsers will work and display OK.  It will confuse a screen reader though.  But nice job - keep it up.

PR #100, opened 2021-06-04
Git checkout b bikes for refugees/haliema 

Haliema26:


SteveLeicester:
Thanks for fixing the bikes for refugees site Haliema.  Good effort.  You have fixed the image and styled the buttons (although not sure they should have a border).  However the event cards on the side do have borders - nice.  There is opportunity for semantic xml - so for example <div class="article">. Very much looks like this should be <article>.  Keep it up Haliema.

PR #102, opened 2021-06-04
northwest 4 Declan Heritage 

Declanhs:


nax2uk (commented):
Hey @Declanhs, good effort on this homework. 😸 

Please read my feedback below especially comments for **Semantic markup** below.


Use semantic markup
Where necessary, replace the <div>s with semantic HTML tags. This will not render differently in the browser, but it's good practice to use semantic markup when you can!
>> You are missing semantic markup. Semantic markup introduces meaning to the web page. Some examples are `<header>`, `<nav>`, `<main>`, `<article>`, `<aside>` and `<footer>`. Have a think of where you can insert these semantic markup in your code.

Fix the broken images
Some of the links to images are broken. The images you need are in the images folder - please replace any broken image links with the correct paths.
>> Fab job in fixing the broken images 🎉
Style buttons
There are 3 buttons on the page. They need to be styled correctly - remember, re-use styles as much as possible by using classes.

Fix the positioning
Use Flexbox to move elements around so they are positioned correctly.
>> The articles for 'Why do refugees need bikes? and 'How can I help?' are vertically stacked instead of horizontal. If you add `<div class='article-wrapper>` on line 103 and adding
```
.article-wrapper {
  display: flex;
  gap: 1rem;
}
```
>> in your css file should fix it.

>> Also the 'Donate Now button and 'Donate a bike today' button should be orangey red.
>> you could add this code to your css file
```
.button--primary {
  background-color: #c05326;
  color: #fff;
}
```

and replace the 'Donate Now' button with this code:
```
<button class="button button--primary">Donate Now</button>
```
you can do the same for the 'Donate a bike today' button

PR #103, opened 2021-06-04
HTML / CSS II

khmdagal:
Khadar M Dagal
Liverpool NW4- HTML / CSS II -Refugee Bikes Project

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -1,14 +1,7 @@
-/**
- * Base styles to use at the start of the class
- *
- * Module: HTML/CSS
- * Lesson: 1,2
- * Class: Scotland 2017
- */
 
-html,
 body {
   font-family: "Source Sans Pro", "Helvetica Neue", Arial, sans-serif;
+
 }
 
 button {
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -67,11 +68,13 @@ button {
 /* Jumbotron */
 
 .jumbotron {
-  background-image: url("../images/header.jpg");
+  background-image: url(/images/header-bike.jpg);
   border-radius: 0;
```
zeinaji, line 72: The picture didn't show for me here, I think because the folder `images` is in a different directory to the file `style.css` (It is in the same directory as its parent folder `styles` however), it needs the `../` at the start.

zeinaji (commented):
This is looking exactly as it should, nice work 🙂

styles/style.css
```diff
@@ -22,22 +15,30 @@ button {
...
+/*I thing I did this above challenge*/
 }
 
 /* Buttons */
-
-/* INSERT BUTTON STYLES HERE */
+.btn-donate, .btn-donate-now {
+  background-color: red;
+  color: white;
+}
```
zeinaji, line 37: Could add a `border-radius` property here to make the button's edges rounder :)

styles/style.css
```diff
@@ -117,10 +128,17 @@ button {
   font-weight: 700;
```
zeinaji, line 128: Could also add a `margin-right` property here to add some space between the pictures and the writing

PR #105, opened 2021-06-08
Module HTML/CSS - Week 2 - Ivan Zhukov - Scotland Class 5

zhkvivan:


sriram-siv:
Good work here Ivan, well done. The only thing that could be improved is the use of semantic tags, I can see a few places where you could use <header> <aside> <footer> etc.
It's good practice for accessibility but it also makes the code much easier to read when you come back to it.

PR #108, opened 2021-06-09
nw4-Gulnihal Naldoken- bikes for refugees

gulnihaln:


styles/style.css
```diff
@@ -67,11 +97,16 @@ button {
 /* Jumbotron */
 
 .jumbotron {
-  background-image: url("../images/header.jpg");
+  background-image: url("/images/header-bike.jpg");
   border-radius: 0;
```
zeinaji, line 101: The picture didn't show for me here, I think because the folder `images` is in a different directory to this file `style.css`, so it needs the `../` at the start

index.html
```diff
@@ -174,8 +181,8 @@ <h3 class="article__title article__title--sidebar">
           <div class="article">
             <img
               class="article__image"
-              src="https://raw.githubusercontent.com/Code-Your-Future/syllabus/master/html-css/assets/news-cycle.jpg"
-              alt=""
+              src="../images/edinburgh-damascus_thumbnail.png"
+              alt="edinburgh-damascus"
```
zeinaji, line 185: Here, images is in the same directory as index.html so it doesn't need the `../`

zeinaji (commented):
Nice work again Gulnihal, it's looking exactly as it should 🙂

styles/style.css
gulnihaln: done

index.html
gulnihaln: done, thanks

PR #121, opened 2021-06-11
bikes-for-refugees-project

TheCheekyGit:
Complete the Bikes For Refugees project


SumanRayamajhi:
Editing unsolves coding

index.html
```diff
@@ -200,7 +205,7 @@ <h3 class="article__title article__title--sidebar">
       <!-- end of row -->
     </div>
 
-    <div class="footer container">
+    <footer class="footer container">
```
alastair87, line 208: Now you've correctly applied semantic HTML ou could fix your styling so as to apply it directly to the `footer` tag. This would mean you'd delete `footer` from the class attribute above and then change the CSS blocks in your CSS file from `.footer` to `footer`. However either version of the code works.

alastair87 (approved):
Excellent work again, you've fulfilled all the major requirements with few issues. 90/100

PR #126, opened 2021-06-12
Bikes for refugees editing some of the code

SumanRayamajhi:
Adding some of the code in HTML and CSS.


SumanRayamajhi:
Hope I did good.

PR #132, opened 2021-07-13
Dilek Canturk - Bikes For Refugees

alastair87:


index.html
```diff
@@ -133,18 +133,18 @@ <h3 class="article__title">How can I help?</h3>
...
           </div>
         </div>
 
-        <div class="sidebar col-lg-4">
+        <aside class="sidebar col-lg-4">
           <h2 class="heading-underline">Upcoming events</h2>
 
-          <div class="article">
+          <article class="article">
             <img
```
alastair87, line 145: You can remove the `class="article"` attribute and target the `article` tag directly in the CSS (change `.article` to article). This applies to all of your `<article>` elements in this document.

index.html
```diff
@@ -195,12 +195,12 @@ <h3 class="article__title article__title--sidebar">
...
           </a>
-        </div>
+        </aside>
       </div>
       <!-- end of row -->
     </div>
 
-    <div class="footer container">
+    <footer class="footer container">
       <div class="row">
```
alastair87, line 204: As with `article` above you could change this just to target the `<footer>` tag without needing to define a footer class (i.e. change `.footer` to `footer` in the CSS).

alastair87:
Excellent work. The page renders as it should and you've made good use of semantic HTML.

alastair87 (commented):
seconds ago
Excellent work. The page renders as it should and you've made good use of semantic HTML.

PR #166, opened 2021-11-13
Added changes to html/css

HamraKhan:
- updated html tag references.
- added/updated css style sheet.
- added learn__more__container styling and referenced it in a specific div tag.
- created css styling for button tags.
- added css class references to button tags.
- updated image tags source and alt attributes.

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -63,10 +63,12 @@ p {
 .navigation__list {
   display: flex;
   list-style: none;
+  margin-left: 30px;;
 }
 
 .navigation__item {
   padding: 0.5rem 0;
+  padding: 15px;
 }
 
 .navigation__link {
@@ -95,7 +97,21 @@ p {
 
 /* Buttons */
 
-  /* INSERT BUTTON STYLES HERE */
+.rect {
+  background-color: var(--orange-dark);
+  color: var(--white);
+  border: none;
+  border-radius: 4px;
+  padding: 5px 12px;
+  margin-left: 32px;;
+}
+
+.white {
+  background-color: var(--white);
+  color: var(--orange-dark);
+  margin-left: 0;
+}
+
 
 
 /* Content */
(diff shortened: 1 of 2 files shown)


LucyMac:
Hi @HamraKhan Can you please edit the name of your PR so that it includes your class and your name please. Look at your classmates's PR names for reference. Thanks 🙂

styles/style.css
```diff
@@ -95,7 +97,21 @@ p {
 
 /* Buttons */
 
-  /* INSERT BUTTON STYLES HERE */
+.rect {
```
ewintram, line 100: Careful with classnames that aren't very descriptive. If we see this classname in the document it's difficult to determine the elements it is being used on, and then it's more difficult to make changes to the class because we aren't sure which elements will be affected. 
You've got the right idea to create some base properties in `.rect` for all buttons, and then override some of these properties in `.white`. Another way to do this more semantically might be to apply the base styles to the button element, e.g.
```
button {
    background-color: var(--orange-dark);
    color: var(--white);
    border: none;
    border-radius: 4px;
    padding: 5px 12px;
    margin-left: 32px;
}
```
and then create a modifier class for the white button, e.g.
```
.button--secondary {
    color: var(--orange-dark);
    margin-left: 0;
}
```
This class can then be used on the white button, if we assume the orange colour is the primary colour, e.g.
```
<button class="button--secondary">Volunteer</button>
```

The `--` is just a naming convention, but semantically we can assume that the new class is used on button elements and will share the same base styles applied to button :)

styles/style.css
```diff
@@ -219,6 +251,17 @@ p {
...
 }
 
+.learn__more__container {
+  display: flex;
+  flex-direction: row;
+  gap: 12px;  
+}
+
+.learn__more__container .article{
+  flex-direction: column;
```
ewintram, line 261: Nice use of flex box

ewintram (commented):
Super job Hamra, it looks just like the picture!

PR #168, opened 2021-11-14
picture and card positions and buttons fixed

mahri09:


mahri09:
London8-Mahri-HTML/CSS-week2

PR #170, opened 2021-11-14
Final

pigos4:
@kolomiets

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,7 +20,7 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header"> 
 
         <a href="#">
           <img
@@ -28,9 +28,10 @@
             alt="Logo for Bikes for Refugees Scotland"
             class="header__logo"
           />
+          
         </a>
 
-        <div>
+        <nav>
           <ul class="navigation__list">
             <li class="navigation__item">
               <a class="navigation__link" href="#">Home</a>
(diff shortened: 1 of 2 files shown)


LucyMac:
Hi @pigos4 Can you please edit the name of your PR so that it includes your class and your name please. Look at your classmates's PR names for reference. Thanks 🙂

index.html
```diff
@@ -45,34 +46,34 @@
...
 
-        <button>Donate Now</button>
+        <button class="red-button">Donate Now</button>
 
-    </div>
+      </header>
 
-    <div class="content">
+    <main class="content">
         <div class="main">
```
kolomiets, line 56: Consider using different [semantic tags](https://www.w3schools.com/html/html5_semantic_elements.asp). Fallback to `div`s when absolutely necessary - quite often there is a better alternative.

index.html
```diff
@@ -20,17 +20,18 @@
...
             class="header__logo"
           />
+          
         </a>
 
-        <div>
+        <nav>
           <ul class="navigation__list">
             <li class="navigation__item">
               <a class="navigation__link" href="#">Home</a>
```
kolomiets, line 37: A stretch goal - what is an alternative way of styling the **nested** tags? Do we really need to apply an explicit style to `a` here or there is another way of handling this in CSS?

styles/style.css
```diff
@@ -75,6 +77,12 @@ p {
   text-transform: uppercase;
   text-decoration: none;
 }
+.navigation__link:hover {
+  color: var(--orange-dark);
+  font-weight: 600;
```
kolomiets, line 82: Do we really need to repeat these styles for `hover`? Can we remove them? Why?

styles/style.css
```diff
@@ -133,7 +143,38 @@ p {
...
+}
+.red-button:hover{
+  color:orangered;
+  background-color: white;
+  border-radius: 5px;
+  border: 0px;
+  padding: 10px;
+}
 
+.white-button{
```
kolomiets, line 161: Consider what can be simplified in you button styles - it feels like a lot of repetition. Try to remove code duplication if possible

kolomiets:
@pigos4 Very good job - the styling of the page look perfect, almost identical to the image provided. 
I've added a couple of comments - mostly about the markup style and code duplication in CSS - there are some things that can be simplified.

Another thing to think about :) If you switch your browser to the mobile view, you'll see that the page looks a bit off. It may be an interesting exercise to figure out why :)

image

PR #172, opened 2021-11-14
Modified HTML and CSS files

SamanZahedi:
I followed the exercises and made some changes in HTML and CSS files
@beckyf0keefe

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,160 +20,165 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header">
+      <a href="#">
+        <img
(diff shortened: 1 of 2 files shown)


beckyf0keefe (commented):
Overall looking really good, just one thing to think about and one stretch goal for if you have time. To think about, where could you use article and section semantic html tags more? Good use of header main sidebar etc. but it would be good to see articles and sections in there. And a stretch goal, currently your div class="article article-sidebar" is wider than the sidebar that surrounds it, this is because of the padding and margin, how could we change it so that we still have the spacing but without the div being wider that the sidebar

PR #174, opened 2021-11-14
London Class 8 - Deago Browne - Bikes for Refugees - Week 2

DeagoBrowne:
Deago Browne

London

Slack: Deago Browne - Trainee

Homework:
Bikes for Refugees
Week 2

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -144,7 +144,6 @@ p {
 .container {
   display: flex;
   flex-direction: column;
-  justify-content: space-around;
 }
 
 .hero h1 {
--- a/styles/style.css
+++ b/styles/style.css
@@ -144,7 +144,7 @@ p {
 .container {
   display: flex;
   flex-direction: column;
-  padding: 60px 0px 20px 10px;
+  padding: 60px 0px 20px 15px;
 }
 
 .hero h1 {
--- a/styles/style.css
+++ b/styles/style.css
@@ -193,6 +193,7 @@ p {
 /* Article */
 .articleContainer {
   display: flex;
+  margin-bottom: 20px;
   
 }
 
@@ -200,7 +201,9 @@ p {
   display:flex;
   border: 2px solid #e4e4e4;
   padding: 15px;
+  padding-right: 20px;
   margin-right: 10px;
+  flex-shrink: 1.1;
 }
 
 .containerTwo {
@@ -208,6 +211,7 @@ p {
   border: 2px solid #e4e4e4;
   padding: 15px;
   margin-left: 10px;
+  
 }
 
 .article {
--- a/styles/style.css
+++ b/styles/style.css
@@ -96,7 +96,7 @@ p {
 /* Buttons */
 .buttonContainer {
   display: flex;
-  padding: 100px;
+  margin-left: 100px;
 }
 
 .buttonBackground {
--- a/styles/style.css
+++ b/styles/style.css
@@ -96,7 +96,8 @@ p {
 /* Buttons */
 .buttonContainer {
   display: flex;
-  margin-left: 100px;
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -95,7 +117,38 @@ p {
...
+.buttonBackground {
+  background-color: #c05326;
+  color: rgb(255, 255, 255);
+  font-weight: 600;
+  border: 6px solid #c05326;
+  border-radius: 4px;
+  margin-left: 33px;
+}
+
+.buttonBackground:hover {
```
mahsa2, line 130: Well done. You've implemented the bonus 👏

styles/style.css
```diff
@@ -95,7 +117,38 @@ p {
 
 /* Buttons */
 
-  /* INSERT BUTTON STYLES HERE */
+
+.buttonBackground {
```
mahsa2, line 121: Extra tip:
When a mouse goes on any clickable item (such as button), we want to tell users to know that's clickable. For that, we convert the cursor into a hand that points to, similar to the link you have. You might want to google `css cursor` to learn more.

styles/style.css
```diff
@@ -209,10 +296,40 @@ p {
...
+}
+
+.fBLink:hover {
+  color: blueviolet;
 }
 
 /* Sidebar */
 
+.sideContainerOne {
+  display:flex;
```
mahsa2, line 309: This won't need a `display: flex` because you are not using any flex property here.

styles/style.css
```diff
@@ -209,10 +296,40 @@ p {
...
+  border: 2px solid #e4e4e4; 
+  padding: 20px 10px 15px 15px;
+}
+
+.sideContainerTwo {
+  display:flex;
```
mahsa2, line 315: Same as above for `flex`.

styles/style.css
```diff
@@ -209,10 +296,40 @@ p {
...
+  border: 2px solid #e4e4e4;
+  padding: 15px 30px 30px 15px;
+}
+
+                                                      /* added flex boxes to readjust the sidebar*/
+.sideContainerThree {
```
mahsa2, line 321: Usually in css we try to use the same css rule in one class for similar items, e.g. sidebar items, articles, etc. And then if any of those items need extra css rules, we add a different class for that one. This way we reuse a class, and all the items in our sidebar list looks almost identical in their common features such as `padding`, `border`, etc.

styles/style.css
```diff
@@ -63,10 +63,28 @@ p {
...
+  margin: 30px;
+}
+
+.nav4 {
+  margin: 30px;
 }
 
 .navigation__item {
-  padding: 0.5rem 0;
+  display: flex;
```
mahsa2, line 86: This won't need a display: flex because you are not using any flex property here.

styles/style.css
```diff
@@ -63,10 +63,28 @@ p {
...
+
+ .nav2 {
+  margin: 30px;
+}
+                                        /* added new classes in order to fix navigation spacing issues */
+.nav3 {
+  margin: 30px;
+}
+
+.nav4 {
```
mahsa2, line 81: In css we can reuse classes in multiple html tags. In this case, I would write only one class like:
```
.nav-item {
    margin: 30px;
}
```
or, because we alreay have a class on all the nav items, I would add `margin` in the existing `.navigation__item` we aleady have:
```
.navigation__item {
     ...
     margin: 30px;
 }
```

In general, in programming we write a lot of code. We want to keep the files as small as possible.

mahsa2 (commented):
Well done, Deago, on finishing this assignment. Congratulations. This is how you learn. Your website looks very nice. 👍 💯

PR #175, opened 2021-11-14
Bikes for refugees

RitaGlushkova:
Made all required changes from exercises 1 - 5. In addition made borders for buttons in hover effects.

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -5,13 +5,13 @@
  * 
  */
 
- :root {
-   --grey-dark: #292b2c;
-   --grey-light: #e4e4e4;
-   --orange-dark: #c05326;
-   --orange-light: #f7eae4;
-   --white: #fff;
- }
+:root {
+  --grey-dark: #292b2c;
+  --grey-light: #e4e4e4;
+  --orange-dark: #c05326;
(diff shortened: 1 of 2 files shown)


RitaGlushkova:
@DiegoJunior0 Hi Henry, I think my bike project is ready to be reviewed:)

styles/style.css
```diff
@@ -201,8 +204,8 @@ p {
   display: flex;
   flex-direction: column;
   width: 45%;
+  justify-content: space-between;
```
RitaGlushkova, line 207: Added to make sure that space is between the articles.

RitaGlushkova, line 204: It didn't work.

DiegoJunior0:
Excellent work Margarita, looks very good.
Well done on getting the button hover working, but be careful doing things on hover that can change size or position (in your case you added a border to the button on hover), as it didn't have one before the hover, this causes the page to flicker a bit when the mouse is over a button.

RitaGlushkova:
@DiegoJunior0 Hi Henry, thank you for your feedback. I added borders to buttons too (same colour as the background of the buttons for the look). Looks like it stopped flickering.

PR #177, opened 2021-11-14
London-class 8-rahwa Ghebremichael-bikes for refugees

rahwafesseha:


mahsa2:
Well done on setting the time aside and making this assignment happen. Congratulations. This is how you learn. Your website looks neat :)Here are a few things that can be improved:

    * This article explains the difference between "aside" and "sidebar". Check if your code is using them right.
    https://aastudio.fr/Sidebar-and-Aside-are-different.html

    * Tag "main" contains the main content only. Sidebar content doesn't stay inside this tag. https://syllabus.codeyourfuture.io/html-css/week-1/lesson
    * Check if index.html line 106 has the correct closing tag.
    * `border: 2 solid rose;` on styles.css (line 78) has CSS console error. You can find it by inspecting your code in the browser and looking for "Console > CSS"
    * The concept of "flexbox" is a bit tough at the beginning. I believe this text has a good explanation that might help to understand it a bit more: https://syllabus.codeyourfuture.io/html-css/week-1/lesson#flexbox. Then, I recommend rechecking if the added flexbox in styles were truly necessary or if any of them are redundant. (e.g. check class '.article__summary')
    * Does class .article__summary need "width: 70%"? Why/Why not? Questions like this are good to ask yourself while programming; they help to understand each concept.

Less important ones:

    * It's a good practice to use "padding" for buttons so that users can have more space to click on them faster and easier especially on the phone.
    * When a mouse goes on any clickable item (such as button), we want to tell users to know that's clickable. For that, we convert the cursor into a hand that points to, similar to the link you have. You might want to google css cursor to learn more.
    * In my opinion, one <section> tag is enough to cover lines 62 to 72, and the inner <section> around buttons looks extra because the content is alreay within a <section> tag.
    * For the bouns parts, you might want to google for css hover examples.

PR #179, opened 2021-11-15
Updated the website

DelroyGayle:
Made changes as described in
https://github.com/CodeYourFuture/bikes-for-refugees

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -42,6 +42,7 @@ p {
   margin: 0 auto;
 }
 
+
 /* Header */
 
 .header {
@@ -56,7 +57,14 @@ p {
 .header__logo {
   max-height: 90px;
   width: auto;
+  object: fill;
 }
+
+.logo__thumbnail {
+  object-fit: content;
+  width: 10rem;
+  margin-right: 1rem;
+} 
 
 /* Navigation */
 
@@ -66,7 +74,7 @@ p {
 }
 
 .navigation__item {
-  padding: 0.5rem 0;
+  padding: 0.5rem 1rem;
 }
 
 .navigation__link {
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -163,8 +206,24 @@ p {
...
+.aPair {
+  display: flex;
+
+
 }
 
+.bottom_articles {
+  display: flex;
+  /* only way I can see to place a suitable gap in between. However there must be another way? */
+  gap: 1rem;
```
amellie, line 224: I actually didn't know this CSS property. I'd use margin on the article itself, but it will leave unwanted gaps on either left or right side. There is another solution using selectors like `first-of-type` and `last-of-type` as can be seen here https://stackoverflow.com/questions/8184958/add-space-between-html-elements-only-using-css but I like your simple solution better!

styles/style.css
```diff
@@ -182,9 +241,11 @@ p {
 }
 
 .article__thumbnail {
-  object-fit: contain;
-  width: 5rem;
-}
+  object-fit: scale-down;
+  /* Lower the width in order to shrink the image */
+  width: 4rem;
+  padding: 0 1rem 0 0;*/
```
amellie, line 247: there is an unexpected `*/` at the end of the line that needs to be removed

styles/style.css
```diff
@@ -202,13 +263,18 @@ p {
   text-decoration: underline;
 }
 
+.article__content{
```
amellie, line 266: empty class, needs to be removed

styles/style.css
```diff
@@ -202,13 +263,18 @@ p {
...
+
 .fa {
   margin-right: 0.5rem;
   color: var(--orange-dark);
 }
 
 .facebook-link {
   color: var(--orange-dark);
+  /* Right-Justify the Logo*/
+  margin-left:12rem;
```
amellie, line 277: I wouldn't use a specific value to right-justify an element. Screen sizes may vary and it may not be 12 rem to right-justify the FB link. Since this is a text, you can use `text-align` on the parent div, or if you still want to use the flexbox, then try playing around with the `justify-content` property

styles/style.css
```diff
@@ -95,8 +103,36 @@ p {
...
+
+button:hover {
+  background-color: #934423;
+}
+
+button:focus {
+  background-color: #934423;
+}
+
+.aButton {
```
amellie, line 116: The two buttons share the similar styles which can be shared. You can have a generic button class which is used in those buttons, and have 2 classes to differentiate between the orange and white ones.

amellie (changes requested):
Hi Delroy, well done on your first CSS assignment. I left some comments here which hopefully can help you to improve your HMTL and CSS. Please do have a go updating those. We can discuss this further next week if you want.

DelroyGayle:
Dear Amalia
Thank you for this

Wow! It's a lot of valuable information.

Lots to learn.

Thank you again 😊

PR #180, opened 2021-11-15
London Class 8 Bikes For Refugees Home-Work Ikaneng Khoza

ikanengKhoza:


index.html
```diff
@@ -100,18 +104,20 @@ <h3 class="article__title">How can I help?</h3>
...
           </div>
         </div>
+       
 
-        <div class="sidebar">
+        <sidebar class="sidebar">
           <h2 class="heading-underline">Upcoming events</h2>
-
+        
           <div class="article">
```
LucyMac, line 117: You can change this `div` to the `article` tag too

index.html
```diff
@@ -47,11 +48,13 @@
...
+        <button class="btn">Donate Now</button>
+
+      </header>
 
-    </div>
 
-    <div class="content">
+    <!--main start-->
+    <main class="content">
         <div class="main">
```
LucyMac, line 58: Best to name the 2nd div `main` and leave the first one as a div. That way, your `aside` further down is not nested inside the `main`.
It's ok for `main` and `aside` to be wrapped inside that `div` with the 'content' class - it might be there for the purposes of CSS. Have you checked what the `.content` class does in the CSS?

styles/style.css
```diff
@@ -120,31 +146,42 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("../images/header-bike.jpg"); /*refers back to the parent folder*/
```
LucyMac, line 149: 👍

styles/style.css
```diff
@@ -159,12 +196,40 @@ p {
...
+    margin: 0px;  
+    padding-top: 2em;
+    
+} 
+
+
+.box {
+  display: flex;
+  flex-wrap: wrap;
+  float: left;
```
LucyMac, line 227: Don't use `float` here you don't need it. Flexbox was created to replace `float` as an easier way to manage layouts. 
What other flexbox rules have you learned that could help with aligning this box to the left?

styles/style.css
```diff
@@ -18,22 +18,39 @@ body {
...
+.btn {
+  background: #c05326;
+  padding: 5px 10px;
+  border-radius: 5px;
+  color:#fff;
+}
+
+.btn_white{
+  background: #fff;
+  color:#c05326;
```
LucyMac, line 44: Nice, good work re-using the `.btn` class and applying the `.btn_white` to modify it for the white button 👍 
Often, these modifiers classes use doubles dashes instead of underscores, like this: `btn--white`. This is a standard defined under the [BEM naming convention](https://css-tricks.com/bem-101/).

styles/style.css
```diff
@@ -18,22 +18,39 @@ body {
 }
 
 body,
-button {
+button,
```
amellie, line 21: Since you use `btn` class for all buttons, you can remove the `button` here

styles/style.css
```diff
@@ -209,6 +276,7 @@ p {
 
 .facebook-link {
   color: var(--orange-dark);
+  margin-left: 11.5em;
```
amellie, line 279: I wouldn't use a specific value to right-justify an element. Screen sizes may vary and it may not be 11.5 em to right-justify the FB link. Since this is a text, you can use text-align on the parent div (which needs to be created)

amellie (commented):
You did very well on this assignment and I have no further comments other than what I have here (which are minimal) and the ones that Lucy have pointed out. Keep up the good work!

styles/style.css
```diff
@@ -103,6 +128,7 @@ p {
 .content {
   display: flex;
   flex-direction: column;
+  margin:0%;
```
amellie, line 131: One more thing, I just saw that your main content is aligned to the left and there are lots of spaces on the right. Try removing this margin and it should work

styles/style.css
ikanengKhoza: removed float

index.html
ikanengKhoza: I changed the 2nd div to main and the site behaved weirdly. I think I'll go back and have a rethink and change this.

index.html
ikanengKhoza: Thanks Lucy, amended

styles/style.css
ikanengKhoza: Thanks Amellie, removed the margin

styles/style.css
ikanengKhoza: Thanks Amellie, I created the div, gave it a class of facebook and aligned right.

PR #181, opened 2021-11-15
applied lots of changes in the index and style files

lalehkoupa:
I have applied lots of changes to my file that my page would be more similar to the sample page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,7 +20,7 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header">
 
         <a href="#">
           <img
@@ -30,7 +30,7 @@
           />
         </a>
 
-        <div>
+        <nav>
           <ul class="navigation__list">
             <li class="navigation__item">
               <a class="navigation__link" href="#">Home</a>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -64,15 +64,15 @@ <h1>Bikes for Refugees</h1>
...
+            <button class="donateBtn">Donate a bike today</button>
+            <button class="volunteerBtn">Volunteer</button>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
-              <div class="article">
+            <div class="articles">
```
Simon2828, line 74:
```suggestion
            <section class="articles">
```
Think maybe a `section` tag is a bit more semantic than a `div` https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div

styles/style.css
```diff
@@ -95,7 +97,29 @@ p {
 
 /* Buttons */
 
-  /* INSERT BUTTON STYLES HERE */
+  button{
```
Simon2828, line 100:
```suggestion
  button {
```
Leave a space between the CSS selector and the bracket. You can update this in a few other places too 🙂

Simon2828 (commented):
Looks really good, Laleh! 😁 Just made a couple of suggestions

index.html
```diff
@@ -100,60 +100,66 @@ <h3 class="article__title">How can I help?</h3>
...
-            <div class="article__content">
-              <h3 class="article__title">
-                <a class="article__title-link" href="#">Join us for our Spring fundraisers</a>
-              </h3>
-              <p class="article__summary">
-                Riders needed now to help raise awareness and funds for BfR and
-                Médecins Sans Frontières.
-              </p>
+          <article class="article">
+            <div class="articles">
```
Simon2828, line 112: You have this the other way round on lines 74-75. Update the `<div>` to a `<section>` which contains the `<article>` tag. 
Check if there's any other places in your html where you could change this.

index.html
lalehkoupa: Thanks @Simon2828 for reviewing my work. I will apply all changes soon.

PR #182, opened 2021-11-15
Bikes for Refugees Dona Shehu

dona-shehu:


index.html
```diff
@@ -30,28 +30,28 @@
...
-        <div>
+        <nav class="nav_bar">
           <ul class="navigation__list">
             <li class="navigation__item">
               <a class="navigation__link" href="#">Home</a>
             </li>
             <li class="navigation__item">
               <a class="navigation__link" href="#">About Us</a>
-            </li>
+          Z  </li>
```
amellie, line 40: You may want to remove this?

styles/style.css
```diff
@@ -95,7 +97,25 @@ p {
 
 /* Buttons */
 
-  /* INSERT BUTTON STYLES HERE */
+  button{
```
amellie, line 100: Instead of using the tag name, use a class name instead. The reason is that, some elements can be represented in different ways using different tags, but they can be the same in terms of styles. Example, `button` and `input[type=submit]` are two elements that are often used to display buttons.

styles/style.css
```diff
@@ -95,7 +97,25 @@ p {
...
+    background-color: var(--orange-dark);
+    color:var(--orange-light);
+    padding:5px;
+    border-radius:5px;
+    margin:-20px 20px 0 -20px;
+  }
+  .button_vol{
```
amellie, line 107: Well-done on not repeating the same styles!

I would suggest a more generic name like `button__primary` and `button__secondary` etc. The reason is that you may want to use this button for other purposes with the same style.

styles/style.css
```diff
@@ -95,7 +97,25 @@ p {
...
+    background-color: white;
+    color:var(--orange-dark)
+    margin-left:0;
+  }
+  button:hover{
+    background-color: var(--grey-dark);
+    
+  }.button_vol:hover{
```
amellie, line 115: I love the hover styles! Can you please make this in a new line, please?

index.html
```diff
@@ -65,13 +65,13 @@ <h1>Bikes for Refugees</h1>
...
-            <button>Volunteer</button>
+            <button class="button_vol"> Volunteer</button>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
+            <div class="learn_more">
               <div class="article">
```
amellie, line 75: You can use an `article` tag for to be more semantic.

Also, there is a lack of spacing between the content of the div and the border (see the screenshot of the homework). Do you know which CSS property that handles this? Please add it :)

index.html
```diff
@@ -104,15 +104,15 @@ <h3 class="article__title">How can I help?</h3>
             </div>
           </div>
         </div>
-
-        <div class="sidebar">
+ 
+        <aside class="sidebar">
           <h2 class="heading-underline">Upcoming events</h2>
 
           <div class="article">
```
amellie, line 111: This can be changed to `article` tag as mentioned above and add more spacing please around the texts.

index.html
```diff
@@ -104,15 +104,15 @@ <h3 class="article__title">How can I help?</h3>
...
             <img
```
amellie, line 112: The image and the text are not aligned as per the homework's screenshot (i.e. image on the left of the text). Would you be able to do that?

index.html
```diff
@@ -163,17 +163,17 @@ <h3 class="article__title">
             <span class="fa fa-facebook-square"></span>
```
amellie, line 163: Facebook link needs to be right-aligned

amellie (commented):
Hi Dona, good job on your first CSS assignment. Can you please check the comments below and see if you can make corrections on those? Thanks very much and let me know if you have any questions!

dona-shehu:
Hi, Amalia! Thank you for reviewing my code.I made the changes that you
suggested.Now after completing  Karma Clone everything makes more sense
😀.I wanted to ask for something that I couldn't find the solution on
my own. At line 199 in Css i applied gap property to create the space
between the articles, the same thing i have done even for the articles in
the aside tag.I tried to applied justify-content but wasn't working.Do you
have any suggestion why it's not working with justify-content?

Thank you,

PR #183, opened 2021-11-15
London Class 8 - Hatice Aydogan - Bikes for Refugees - Week 2

Hatice06:
London Class 8- Hatice Aydogan -Bikes for Refugees - Week 2

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,7 +30,7 @@
           />
         </a>
 
-        <nav>
+        <nav class="nav-bar">
           <ul class="navigation__list">
             <li class="navigation__item">
               <a class="navigation__link" href="#">Home</a>
@@ -64,7 +64,7 @@
               seekers in Scotland.
             </p>
             <button>Donate a bike today</button>
-            <button>Volunteer</button>
+            <button class="volunteer-btn">Volunteer</button>
           </div>
 
           <div>
@@ -110,8 +110,8 @@
           <article class="article">
             <img
               class="article__thumbnail"
-              src="spring-fundraisers_thumbnail.jpg"
-              alt=""
+              src="./images/spring-fundraisers_thumbnail.jpg"
+              alt="Spring fundraisers image blooming on branch with lettering logo"
             />
             <div class="article__content">
               <h3 class="article__title">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -45,34 +45,33 @@
...
 
         <button>Donate Now</button>
 
-    </div>
+      </header>
 
     <div class="content">
-        <div class="main">
-          <div class="alert">
+        <main class="main">
```
ewintram, line 55: Very nice use of semantic elements!

index.html
```diff
@@ -45,34 +45,33 @@
...
+            <button class="volunteer-btn">Volunteer</button>
           </div>
 
-          <div>
+          <div class="container-article-main">
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
-              <div class="article">
+            <div class="article-main">
```
ewintram, line 73: Good descriptive class names

styles/style.css
```diff
@@ -24,6 +24,28 @@ button {
...
 
+button {
+  padding: 0.4rem 0.6rem;
+  background-color: var(--orange-dark);
+  border: none;
+  color: var(--white);
+  border-radius: 0.2rem;
+  font-weight: 600;
+}
+button:hover {
```
ewintram, line 35: Well done on adding the hover effects!

styles/style.css
```diff
@@ -24,6 +24,28 @@ button {
...
+  background-color: rgb(110, 108, 218);
+}
+
+.volunteer-btn {
```
ewintram, line 39: Well done on creating this variation on the button! 

Something you could try next time is to use a CSS naming convention for creating these kind of modifier classes. You use the name of the root class, followed by `--`, followed by the name of the modifier, e.g. in your case
```
.button--volunteer {
    color: var(--orange-dark);
}
```

Then imagine in the future if you wanted to add another class for a purple donate button, you could create
```
.button--donate {
    background-color: var(--purple);
    color: var(--orange-dark);
}
```
This is just a naming convention, but semantically we can assume that your classes are used on button elements and will share the styles applied to button :)

index.html
```diff
@@ -45,34 +45,33 @@
...
+              <article class="article article-main-content">
```
ewintram, line 74: Watch out when you use multiple classes on one element (e.g. `article` and `article-main-content`) in case you duplicate styles in these two classes (e.g. both of these classes set `display: flex`, but you only need to set this property once)

ewintram (commented):
Amazing job Hatice! The webpage looks perfect

PR #185, opened 2021-11-15
London Class 8 - ANDERSON FERREIRA - Bikes for Refugees - Week 2

andersonuk:


index.html
```diff
@@ -64,15 +64,15 @@ <h1>Bikes for Refugees</h1>
...
+            <a class="button" href="#">Donate a bike today</a>
+             <a class="button" href="#">Volunteer</a>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
-              <div class="article">
+            <section  class="article">
```
ewintram, line 74: Great use of semantic elements!

index.html
```diff
@@ -100,18 +100,19 @@ <h3 class="article__title">How can I help?</h3>
...
-              </div>
-            </div>
+              </article>
+            </section>
           </div>
-        </div>
 
-        <div class="sidebar">
+        </div>
+          <!-- </main> -->
```
ewintram, line 108: Don't forget to remove commented out code

index.html
```diff
@@ -47,11 +47,11 @@
           </ul>
         </div>
 
-        <button>Donate Now</button>
+        <a class="button donation" href="#">Donate Now</a>
```
ewintram, line 50: Interesting that you chose to use a link (anchor element) for the buttons. Links should indeed be used for navigating to another page, however it's confusing to users if a link is styled to look like a button. So in this case, I think the original design could be improved :)

index.html
```diff
@@ -141,7 +142,7 @@ <h3 class="article__title">
           <div class="article">
             <img
               class="article__thumbnail"
-              src="edinburgh-damascus_thumbnail.png"
+              src="../images/edinburgh-damascus_thumbnail.png"
```
ewintram, line 145: These images aren't displaying for me, can you fix them?

styles/style.css
```diff
@@ -95,8 +104,38 @@ p {
...
+    text-decoration: none;
+    display: inline-block;
+    border-radius: 3px;
+    border:none;
+    padding: 8px 12px;
+    color:var(--white);
+    text-decoration: none;
+
+}
+  .button:hover {
```
ewintram, line 119: Great use of the hover effect!

styles/style.css
```diff
@@ -95,8 +104,38 @@ p {
...
+    background-color: var(--white);
+    text-align: center;
+    text-decoration: none;
+    border-radius: 3px;
+    padding: 8px 12px;
+    color:var(--orange-dark);
+
+}
+
+  .donation:hover {
```
ewintram, line 129: There are some duplicated properties in your button classes. Do you know how you can make this code more DRY (Don't Repeat Yourself)? Let me know if you would like an example :)

styles/style.css
```diff
@@ -163,12 +206,26 @@ p {
...
 }
 
 .article__title {
   margin-bottom: 0.5rem;
   font-size: 1rem;
   font-weight: 700;
+
+}
+.article__content {
+padding: 9px 9px;
```
ewintram, line 227: Some of the indentation in this file isn't quite right. If you install the Prettier extension in VS Code and enable 'Format on Save' in your VS Code settings it should fix all the indentation for you. Let me know if you need any help!

ewintram (commented):
Great job Anderson! It looks great

index.html
andersonuk: I did it just to show it supposed to be a link :)

index.html
andersonuk: I use to create comments for the parents close tags so I can't read my code easily. 
Is not that a good pratice?

index.html
andersonuk: I updated and wrote alt's as forgot before :)
Let me know if the image is working.

One question, please. Do you have a tool where you can check the website without downloading it to your computer?

styles/style.css
andersonuk: Your right! I've  cleaned up about 7 lines :)

styles/style.css
andersonuk: Your right! I've cleaned up about 7 lines :)

index.html
ewintram: Comments in HTML tend to be confusing. I would say the purpose of the semantic tags is that they show you where your elements open and close, so you shouldn't need comments too

index.html
ewintram: Great that you added the `alt` attributes! The images aren't displaying for me, because the file path is `./images`, instead of `../images`, because the index.html is in the same directory as the `images` folder

PR #190, opened 2021-11-16
styles by Monika Dangol

Monika39-maker:
please check

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,7 +20,7 @@
   </head>
 
   <body>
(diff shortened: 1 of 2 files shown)


beckyf0keefe (commented):
Overall getting there, just needs some more padding to get it a little closer to the original design but good use of semantic html. Try using flex-box on the learn more section as well as on the hero and it will be easier than trying to add padding and margin to the hero

PR #193, opened 2021-11-16
LONDON CLASS 8 - ALI NURISIAHSIAHI - HTML/CSS- BIKES FOR REFUGEES - WEEK2

AliNurisiahsiahi:
CHANGED

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,7 +66,7 @@
               seekers in Scotland.
             </p>
             <button>Donate a bike today</button>
-            <button class="buttonNew">Volunteer</button>
+            <button>Volunteer</button>
           </div>
 
           <div>
--- a/styles/style.css
+++ b/styles/style.css
@@ -66,7 +66,8 @@ p {
 }
 
 .navigation__item {
-  padding: 0.5rem 0;
+  padding: 0.5rem 0.8rem;
+  
 }
 
 .navigation__link {
@@ -75,6 +76,9 @@ p {
   text-transform: uppercase;
   text-decoration: none;
 }
+.navigation__link:hover{
+  color: #c05326;
+} 
 
 
 /* Text styles */
@@ -101,12 +105,9 @@ button {
   border: none;
   border-radius: 4px;
 }
-.buttonNew{ 
+button:hover{ 
   color:var(--orange-dark);
   background-color: white;
-  padding: 0.4rem 0.6rem;
-  border: none;
-  border-radius: 4px;
 }
   /* INSERT BUTTON STYLES HERE */
 


index.html
```diff
@@ -49,16 +49,17 @@
 
         <button>Donate Now</button>
 
-    </div>
+    </header>
 
     <div class="content">
```
kolomiets, line 54: Consider using different [semantic tags](https://www.w3schools.com/html/html5_semantic_elements.asp) across the page. Fallback to `div`s when absolutely necessary - quite often there is a better alternative.

index.html
```diff
@@ -49,16 +49,17 @@
...
-        <div class="main">
+        <main class="main">
           <div class="alert">
             You've helped us give <span class="text-highlight">72</span> bikes
             to refugees and asylum seekers so far. <strong>Thank you!</strong>
           </div>
 
           <div class="hero">
+            
             <h1>Bikes for Refugees</h1>
```
kolomiets, line 63: Please take a look at the buttons on the target image - can we make them look different? (one is white, another is brown/orange)

index.html
```diff
@@ -100,18 +101,18 @@ <h3 class="article__title">How can I help?</h3>
                     Learn more
                   </a>
                 </p>
-              </div>
+              </article1>
```
kolomiets, line 104: typo, should be `article`?

styles/style.css
```diff
@@ -94,7 +98,17 @@ p {
...
+  background-color:var(--orange-dark);
+  padding: 0.4rem 0.6rem;
+  border: none;
+  border-radius: 4px;
+}
+button:hover{ 
+  color:var(--orange-dark);
+  background-color: white;
+}
   /* INSERT BUTTON STYLES HERE */
```
kolomiets, line 112: Should we move button style under this comment? Or remove the comment altogether. Now it is quite confusing for the reader :)

kolomiets:
Great work @AliNurisiahsiahi ! The page looks very close to the targe image - very well done!

I've added a couple of comments - please take a look. We can discuss further if anything is not clear. 

An additional advanced problem to think about - if you switch your browser to mobile mode, you'll see that the page looks a bit off. Think about what could be the reason?

image

styles/style.css
AliNurisiahsiahi: Yes

index.html
AliNurisiahsiahi: thank you, I have done it.

AliNurisiahsiahi:
> Great work @AliNurisiahsiahi ! The page looks very close to the targe image - very well done!
> 
> I've added a couple of comments - please take a look. We can discuss further if anything is not clear.
> 
> An additional advanced problem to think about - if you switch your browser to mobile mode, you'll see that the page looks a bit off. Think about what could be the reason?
> 
> image

thank you, I think you can review it. now it's much better.

PR #196, opened 2021-11-16
bike homework

mairacagri:
setting your html and css close to the example

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,7 +20,8 @@
   </head>
 
   <body>
-    <div class="header">
+    <!-- chaanged <div class ="header" > to <header> tag-->
+      <header class="header">
 
         <a href="#">
           <img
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -46,32 +47,34 @@
...
             to refugees and asylum seekers so far. <strong>Thank you!</strong>
           </div>
 
           <div class="hero">
+            <!-- still looking how to add background bike photo , nevermind founded :)-->
             <h1>Bikes for Refugees</h1>
             <p>
               Providing donated bikes and accessories to refugees and asylum
-              seekers in Scotland.
+              seekers in England.
```
mahsa2, line 68: Good catch :D I don't think it was intended though. Being detailed is a good asset in programming.

index.html
```diff
@@ -103,18 +106,19 @@ <h3 class="article__title">How can I help?</h3>
               </div>
             </div>
           </div>
-        </div>
+        </main>
 
-        <div class="sidebar">
+        <sidebar class="sidebar">
```
mahsa2, line 111: `sidebar` is different from `aside`. 
In short, `aside` is a html tag but `sidebar` is just the name of an element in the website layout.
Please read more here: https://aastudio.fr/Sidebar-and-Aside-are-different.html

index.html
```diff
@@ -46,32 +47,34 @@
...
+            <button class="donate_now">Donate a bike today</button>
+            <button class="volunteer">Volunteer</button>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
-
-            <div>
+           <!-- created parent element gave class to this elemet -->
+            <div class="flexbox">
```
mahsa2, line 77: It's usually better to pick a class name for css classes that describes the content your user will see. E.g. here you could say `class=articles` instead of `class=flexbox`.

styles/style.css
```diff
@@ -20,8 +20,34 @@ body {
...
+  
+  color:#292b2c;
   font-size: 1rem;
+  /* donate now button */
+  border-radius: 5px;
+  
+
+}
+/* changing 2 bottons donate now and donate bike today */
+.donate_now {
```
mahsa2, line 32: It's great that you reused this class for the two similar buttons.
Another thing that is common in programming is to combine all the features of similar buttons in one class, let's say '.general-button' with all the mutual css rules in `.donate_now` and `.volunteer`. Then we add a second class on the button to define the different css rules like the `color` or `margin-left`.

mahsa2: Also, when a mouse goes on any clickable item (such as button), we want to tell users to know that's clickable. For that, we convert the cursor into a hand that points to, similar to the link you have. You might want to google css cursor to learn more.

mahsa2 (commented):
Maira, well done on finishing this assignment. Congratulations. This is how you learn. Your website looks very nice. Beautiful 👍 💯

index.html
mairacagri: Thank you Mahsa for your feedback. i checked the link you send me and i understood what i did wrong

styles/style.css
mairacagri: Yes hand point cursor  i completely  missed it will definitely use it in Karma assignment. Thank you 👍

PR #197, opened 2021-11-16
London class8/BusraErdogan/Bikes for Refugees/Week2 

BusraElmasErdogan:
CSS Updates

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,157 +20,162 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header">
+      <a href="#">
+        <img
+          src="images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees England"
+          class="header__logo"
+        />
+      </a>
 
-        <a href="#">
-          <img
-            src="images/bikes-for-refugees_logo.jpg"
-            alt="Logo for Bikes for Refugees Scotland"
-            class="header__logo"
-          />
-        </a>
+      <nav>
+        <ul class="navigation__list">
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Home</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">About Us</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">News & Events</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </nav>
+
+      <button class="dark-button">Donate Now</button>
+    </header>
+
+    <section class="content">
+      <main class="main">
+        <div class="alert">
+          You've helped us give <span class="text-highlight">72</span> bikes to
+          refugees and asylum seekers so far. <strong>Thank you!</strong>
+        </div>
+
+        <div class="hero">
+          <h1>Bikes for Refugees</h1>
+          <p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -20,157 +20,162 @@
...
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </nav>
+
+      <button class="dark-button">Donate Now</button>
+    </header>
+
+    <section class="content">
```
kolomiets, line 52: Great to use semantic tag `section` here, but what about all other `div` tags on the page? Would you be able to replace there generic tags with something more meaningful? Use [semantic tags](https://www.w3schools.com/html/html5_semantic_elements.asp) for inspiration

index.html
```diff
@@ -20,157 +20,162 @@
...
-              class="article__thumbnail"
-              src="spring-fundraisers_thumbnail.jpg"
-              alt=""
-            />
-            <div class="article__content">
-              <h3 class="article__title">
-                <a class="article__title-link" href="#">Join us for our Spring fundraisers</a>
-              </h3>
+            <div class="article">
+              <h3 class="article__title">How can I help?</h3>
```
kolomiets, line 89: A stretch goal - what is an alternative way of styling the **nested** tags? Do we really need to apply an explicit style to h3 here or there is another way of handling this in CSS?

index.html
```diff
@@ -20,157 +20,162 @@
...
+          <img
+            class="article__thumbnail"
+            src="images/spring-fundraisers_thumbnail.jpg"
+            alt=""
+          />
+          <div class="article__content">
+            <h3 class="article__title">
+              <a class="article__title-link" href="#"
+                >Join us for our Spring fundraisers</a
+              >
```
kolomiets, line 119: Formatting is a bit unusual here. It's a matter of personal style of course, but I'd suggest to keep `>` on the same line with `<`

styles/style.css
```diff
@@ -24,7 +24,12 @@ button {
   font-size: 1rem;
 }
 
-h1,h2,h3,h4,h5,h6 {
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
```
kolomiets, line 32: Are we planning to use headers up to 6th level? :)

index.html
```diff
@@ -20,157 +20,162 @@
...
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">News & Events</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </nav>
+
+      <button class="dark-button">Donate Now</button>
```
kolomiets, line 49: Review the button styles - it looks we have a border around them that shouldn't be there, according to the target image on GitHub

styles/style.css
```diff
@@ -92,11 +98,22 @@ p {
...
+  background-color: var(--orange-dark);
+  color: var(--white);
+  border-radius: 0.5rem;
+  padding: 0.5rem;
+}
 
+.light-button {
+  background-color: var(--white);
+  color: var(--orange-dark);
+  border-radius: 0.5rem;
```
kolomiets, line 113: Looks like we use the same settings for border radius and padding for both button types. Can we reduce code duplication?

kolomiets:
Great solution - the design looks almost identical to the image provided, very good job! I've left a couple of comments/suggestions - please take a look and we can discuss if something is not clear.

PR #203, opened 2021-11-18
Initial commit

LovesPictures:
Week 1 coursework


styles/style.css
```diff
@@ -74,10 +81,18 @@ p {
...
   text-transform: uppercase;
   text-decoration: none;
+  padding: 0.4rem 1rem 0.4rem 1rem;
 }
 
+/*: Bonus: hover effect: top menu navigation become orange on hover 
+The colours of the buttons are inverted on hover */
+a:hover {
+  background-color: var(--orange-dark);
+  color: var(--white);
```
Simon2828, line 91: When you hover over here: 
image
the text "Learn more" from line 79 in the index.html:
```
<a href="#">
  <span class="arrow-right"></span>
  Learn more
</a>
```
 is not showing as white... can you work out why?

Simon2828 (commented):
This is really great work! 😄

styles/style.css
LovesPictures: Hello Simon, 

It's updated now: I made the amendments to  275 - 289
    `/* Link before */
    .article__read-more a {
      font-size: 0.85rem;
      font-weight: 700;
      text-decoration: none;
      color: var(--orange-dark);
    }
    /* link during and afterwards */
    .article__read-more a:hover,
    .article__read-more a:focus {
      text-decoration: none;
      color: var(--white);
      font-weight: 400;
      padding-right: 1rem;
    }`

PR #204, opened 2021-11-18
Some updates added

ShadiSafaee:


index.html
```diff
@@ -21,156 +21,162 @@
...
-                Médecins Sans Frontières.
+                Many refugees are placed in housing in areas where there are few
+                jobs. Bikes provide low-cost transportation so that they can get
+                to work, manage child care and do their shopping.
+              </p>
+              <p class="article__read-more">
+                <a href="#">
+                  <span class="arrow-right"></span>
+                  Learn more
+                </a>
```
beckyf0keefe, line 84: Overall good use of semantic html, but where do you think we could use the article tag in here

styles/style.css
```diff
@@ -22,13 +22,22 @@ button {
...
+h4,
+h5,
+h6 {
   margin-top: 0;
   margin-bottom: 0;
 }
 
+h2 {
+  margin-top: 60px;
+}
```
beckyf0keefe, line 40: How do you think we could apply this to the learn more h2 _without_ adding it to the uncoming events h2? So that we wont have the gap at the top sidebar

beckyf0keefe (commented):
Overall looking really good, just one or two things for you to think about. An extra goal is for you if you have time is to add the onHover to the buttons and the menu items in the nav bar.

PR #253, opened 2022-06-30
WM4-Vanipriya_Piramanathan-bike_for_refugees-Week2

Vanipriyaarun:


jcholyhead (commented):
Good use of semantic html. Good reproduction of the design in the spec.

Recommendation for future: use more meaningful class names. "button" and "button2" are not descriptive and using generic names like this can increase the potential for name clashes in more complicated sites.

PR #261, opened 2022-07-01
modifications

FranklinPTorres:


jcholyhead (commented):
Good attempt. 
You've missed the borders around the articles on the right hand side of the page, but that a minor aesthetic thing.

The divs with class="article" would be more semantic if they were article elements.

PR #323, opened 2022-09-13
 Exercise for HTML/CSS - Lesson 1

JustinasBrazauskas:


styles/style.css
```diff
@@ -162,39 +202,65 @@ p {
 /* Article */
 
 .article {
```
magarpratik, line 204: Could you position the two articles side by side like the picture from the assignment brief? Using flexbox may help here.
Screenshot 2022-09-14 at 23 02 05

magarpratik (commented):
Good effort so far!
Please make the requested changes.

PR #330, opened 2022-10-23
London 9 - Group 1 - Natalie Zablotska - HTML & CSS

nataliiazab:
Your Name: Natalie Zablotska
Your City: London
Your Slack Name: Natalie
Homework Details
Module: HTML/CSS
Week: 1

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,7 +47,7 @@
           </ul>
         </div>
 
-        <button>Donate Now</button>
+        <button class="btn1">Donate Now</button>
 
     </div>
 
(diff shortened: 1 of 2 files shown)


Hamisakim:
Nice work  😀 


for the images, you need to make sure the index.html knows the relative file path to those images. i.e 
RIGHT : "./images/edinburgh-damascus_thumbnail.png"
WRONG : "edinburgh-damascus_thumbnail.png"


Here's a tip, when you add a border it adds those to the element's size, which is why the buttons move a little when hovering. Try using `outline`  instead. [more here](https://www.w3schools.com/css/css_outline.asp#:~:text=Note%3A%20Outline%20differs%20from%20borders,the%20width%20of%20the%20outline.)

PR #334, opened 2022-10-24
London9 - Mo Nahvi- HTML/CSS - week1

monahvi:


index.html
```diff
@@ -64,15 +66,15 @@ <h1>Bikes for Refugees</h1>
...
+            <button class="secondBS">Volunteer</button>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
-              <div class="article">
+            <section>
+              <article>
```
nuclear-pasta, line 77: It would be useful to use a class also here, so that in the CSS you can define the ratio of flexbox.
Try to change here the `flex` attribute on children and see what happens:
https://www.w3schools.com/cssref/tryit.php?filename=trycss3_flex

if you use the CSS directly on article element , could affect also other `article` element, while we want apply this change only to this article element and the next one

styles/style.css
```diff
@@ -5,246 +5,310 @@
...
+ margin-bottom: 2rem;
 }
 
+
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
-  background-color: var(--grey-light);
+ background-image: url("/images/header-bike.jpg");
```
nuclear-pasta, line 151: in order to retrieve the background image, try to start from the location of the file you are working on:
```
styles/style.css
```
and how from there you would get to the file you need:

```
images/header-bike.jpg
```
In this case you would need to go a level up and then retrieve the image.

https://www.ibm.com/docs/en/zos/2.1.0?topic=names-dot-notation

https://www.loom.com/share/69f1549e89a740b8864c11c2b85a5d66

index.html
```diff
@@ -100,19 +102,20 @@ <h3 class="article__title">How can I help?</h3>
...
+        <aside>
           <h2 class="heading-underline">Upcoming events</h2>
 
-          <div class="article">
+          <article>
             <img
               class="article__thumbnail"
-              src="spring-fundraisers_thumbnail.jpg"
-              alt=""
+              src="/images/spring-fundraisers_thumbnail.jpg"
```
nuclear-pasta, line 117: same thing
( https://github.com/CodeYourFuture/bikes-for-refugees/pull/334/files#r1003856170 )
applies here

PR #343, opened 2022-10-27
London9 - HTML/CSS - week1

hkuhku609:
finished

### DIFF
--- a/index.html
+++ b/index.html
@@ -56,7 +56,7 @@
           refugees and asylum seekers so far. <strong>Thank you!</strong>
         </div>
 
-        <div class="hero">
+        <section class="hero">
           <h1>Bikes for Refugees</h1>
           <p>
             Providing donated bikes and accessories to refugees and asylum
@@ -64,7 +64,7 @@
           </p>
           <button class="brownbtn">Donate a bike today</button>
           <button>Volunteer</button>
-        </div>
+        </section>
 
         <div>
           <h2 class="heading-underline">Learn more</h2>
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -248,3 +248,20 @@ p {
...
 }
+
+button {
+  background-color: var(--white);
+  border: 1px solid transparent;
+  border-radius: 8px;
+  padding: 0.5rem;
+  color: var(--orange-dark);
+  font-weight: bolder;
+}
```
samhoooo, line 275: It would be a good practice to define the appearance of the Volunteer button with a **new css class**, instead of defining it only on button element.

This can ensure the definition do not affect other button elements, such as  `.brownbtn`, and reduce conflicts when you have more types of buttons.

To improve, we can extract similarity across buttons and generalise it on the button element. For example:

```
button {
   border: 1px solid transparent;
   border-radius: 8px;
   padding: 0.5rem;
   font-weight: bolder;
}
.whitebtn {
  color: var(--orange-dark);
}
```

index.html
```diff
@@ -62,7 +62,7 @@ <h1>Bikes for Refugees</h1>
             Providing donated bikes and accessories to refugees and asylum
             seekers in Scotland.
           </p>
-          <button>Donate a bike today</button>
+          <button class="brownbtn">Donate a bike today</button>
           <button>Volunteer</button>
```
samhoooo, line 66: It is better to give a class here

samhoooo (commented):
Great work! Here are some suggestions for you

styles/style.css
hkuhku609: wow, this is the best idea, thank you for suggestion

PR #348, opened 2022-10-27
bikes-for-refugees

KristinaDudnyk:
1) Use semantic markup
2) Fix the broken images
3) Style buttons
4) Add spacing
5) Fix the positioning
Bonus: hover effects


KristinaDudnyk:
.

PR #355, opened 2022-10-28
Mickey Haile london9 coursework week 2 

mickeyhaile2:


index.html
```diff
@@ -125,24 +125,24 @@ <h3 class="article__title">
             </div>
           </div>
 
-          <div class="article">
+          <article class="article">
```
JDysiewicz, line 128: nice use of semantic html :)

index.html
```diff
@@ -125,24 +125,24 @@ <h3 class="article__title">
             </div>
           </div>
```
JDysiewicz, line 126: this closing `</div>` tag needs to be a `</article>` tag to match the above; this change will fix a lot of the styling issues on the site
image
image

styles/style.css
```diff
@@ -75,6 +102,17 @@ p {
...
+  color: var(--orange-dark)
+}
+.orange__button:hover {
+  background-color: var(--white);
+  color: var(--orange-dark);
+}
+.white__button:hover {
+  background-color: var(--orange-dark);
+  color: var(--white);
+}
```
JDysiewicz, line 115: Nice use of button styling! :)

styles/style.css
```diff
@@ -122,6 +160,9 @@ p {
 .hero {
   background-image: url("header-bike.jpg");
```
JDysiewicz, line 161: This image link is still broken; currently it is looking for the image in `/styles/header-bike.jpg`, however the image is in the `/images` folder. There are two ways you can fix this link; either by _relative_ import ("what is the path to get to the destination from where I am currently") or _absolute_ ("starting at the very top level, what is the path to the image").

JDysiewicz: This goes for the other images on the page too; although they are in the HTML using the `src` attribute, the same logic applies.

JDysiewicz (changes requested):
The site is really getting there! There's just a couple things that I'd like you to fix:
1. The image imports such that the images are displayed correctly
2. The sidebar styling (see comment about a missing `</article>` closing tag
3. The styling of the "Learn More" section; these need to be displayed side-by-side with some padding between them. A good place to start is applying the `display: flex` property to the parent `<div>` tag on line 72.

PR #359, opened 2022-10-28
london-class9-lovelace_Meysam_Arshadi_HTML_CSS_WEEK_02

Meysam-Arshadi:


styles/style.css
```diff
@@ -120,25 +123,36 @@ p {
...
+  justify-content: space-between;
+}
+
+.header-bike {
+  position: relative;
 }
 
 .hero h1 {
-  margin-bottom: 1rem;
+  position: absolute;
```
nuclear-pasta, line 136: https://www.loom.com/share/77e00645bcd14c4296e0f8992378be48

index.html
```diff
@@ -20,160 +20,171 @@
...
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </nav>
+
+      <button class="button-orange">Donate Now</button>
+    </header>
+
+    <section class="content">
+      <main class="main">
```
nuclear-pasta, line 53: I would try to give a height to this element. Delete the  tag from inside and use it as background image and then use flexbox the way you were using it

styles/style.css
```diff
@@ -120,25 +123,36 @@ p {
...
+  top: 35%;
+  margin-left: 40px;
 }
 
-
 /* Headings */
 
 .heading-underline {
   position: relative;
+  margin-top: 30px;
```
nuclear-pasta, line 155: try to be consistent using `rem` instead of pixel.
rem is preferred as it is more responsive 
Here a reading about it
https://www.digidop.fr/en/blog/rem-pixel-what-a-difference-site

PR #360, opened 2022-10-28
LONDON-CLASS-9_SAGHAR-HOSSEINMARDI-HTML/CSS-WEEK-2

SagharHosseinmardi:
I could not submit my work as i wanted it looks like, the Flex box and also HTML tags and finding the written classes and recognise them in CSS was very hard for me.

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -5,13 +5,13 @@
  * 
  */
 
- :root {
-   --grey-dark: #292b2c;
-   --grey-light: #e4e4e4;
-   --orange-dark: #c05326;
-   --orange-light: #f7eae4;
-   --white: #fff;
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -120,21 +124,24 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("../images/header-bike.jpg");
   background-color: var(--grey-light);
+  margin-bottom: 3rem;
+  padding: 80px 40px;
```
jonnywyatt, line 130: Does the job!

jonnywyatt (commented):
Looks good!

PR #368, opened 2022-10-28
London9_Poliakova_Inna_Bikes_for_refugees

Innapoliakova:


Abdirizak-Mohamed:
Hey Inna, 

I really liked this submission. The images all appear correctly and while the sidebar is not in position the page still looks really good! 

To resolve the sidebar issue, I would look either @ using [bootstrap grid](https://getbootstrap.com/docs/5.2/layout/grid). Or alternatively maybe you could add a parent element that has flex and place the content and aside in this row div.

PR #382, opened 2023-01-20
adding colour to button

SalihaPopal:


SalihaPopal:
added semantic markup, fixed broken img, .......week2

PR #387, opened 2023-01-21
Completed the required HTML and CSS Bike for Refugees exercise for Week1

Afsha10:
Made necessary changes to exercise 1 and exercise 2. And also did the exercises to change the left box background to grey and right box border. Made changes in padding around the "Bike for Refugees" to 3rem, margin around the buttons to 0.25rem.

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -22,6 +22,7 @@ button {
   font-family: "Source Sans Pro", "Helvetica Neue", Arial, sans-serif;
   color: var(--grey-dark);
   font-size: 1rem;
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -120,18 +171,50 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("../images/header-bike.jpg");
   background-color: var(--grey-light);
```
jxz12, line 175: here you can try `background-size: 75%` and it should look more similar to the final design

PR #403, opened 2023-01-28
changes on navbars

85pouya:
I have made some changes on navbars (class="navigation__link), in html file.


85pouya:
I have changed the position of navigation bars.
I also placed the image behind the text and resized the alert.

PR #451, opened 2023-08-09
Abdihamid Ali - WM5 - Bikes-for-Refugees

AbdihamidAli:
This is my HTML-CSS/sprints 1/prep bikes-for-refugees task completed

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,179 +1,134 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+<head>
+  <!-- Meta tags for character set, viewport, and compatibility -->
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
 
-    <link rel="dns-prefetch" href="//fonts.googleapis.com" />
-    <link
-      href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600,700,"
-      rel="stylesheet"
-    />
-    <link
-      rel="stylesheet"
-      href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
-    />
-    <link rel="stylesheet" href="styles/style.css" />
+  <!-- Prefetch fonts and link to external font and icon stylesheets -->
+  <link rel="dns-prefetch" href="//fonts.googleapis.com" />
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1,179 +1,134 @@
...
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
+
+  <!-- Link to the stylesheet -->
+  <link rel="stylesheet" href="styles/style.css" />
+
+  <!-- Title of the webpage -->
+  <title>Bikes for Refugees</title>
+</head>
+<body>
+  <!-- Header and Navigation -->
```
MarcusZagorski, line 21: Could I ask the reasoning behind your use of comments please? I think maybe you should minimise them. Comments should be used in specific sections

index.html
```diff
@@ -1,179 +1,134 @@
...
-                <p class="article__read-more">
-                  <a href="#">
-                    <span class="arrow-right"></span>
-                    Learn more
-                  </a>
-                </p>
-              </div>
-            </div>
-          </div>
+          <button class="donate-btn">Donate a bike today</button>
```
MarcusZagorski: Do you think maybe it would be better to use anchors instead of buttons? As its still possible to style them to look like buttons, as these links are used to direct you to other pages

index.html
MarcusZagorski, line 1: Great work Abdi! I think the only other thing that I can include regarding your HTML on a whole, is to add semantic HTML. If you do not understand what this is a quick Google search will tell you everything you need to know

MarcusZagorski (commented):
Good work Abdi. I would also advise that you remove the hover drop shadow from your CSS.

index.html
AbdihamidAli: @MarcusZagorski, Mostly I use the comments to differentiate between the sections within the body tag. That's why I use comments like a title before going to the next section. Maybe I'm wrong doing this, but it helps me organized while coding.

index.html
AbdihamidAli: sure, thank you for pointing this out... I've made the changes :1st_place_medal:

index.html
AbdihamidAli: Many thanks for your comments, Yes went to Google and looked into semantic HTML "
semantic HTML elements like header, nav, main, article, section, aside, and footer are used to structure the webpage's content in a meaningful way. Each element represents a specific section and purpose, making it easier to understand the layout and purpose of different parts of the webpage."

 I implemented them into my webpage...
 you are amazing for pointing out these improvements. 
 I appreciate you :pray:

index.html
MarcusZagorski: @AbdihamidAli yes this is absolutely fine, but what I was trying to say is that you could minimise the amount of comments you use and group sections more accordingly, my apologies if I wasn't clear when I first commented.

PR #452, opened 2023-08-09
WM5 | Adniya Yousaf | Module MTML-CSS | Week 1

adniyaYousaf:


styles/style.css
```diff
@@ -120,18 +162,25 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  display: block;
```
MarcusZagorski, line 165: Why do you think that display: block; was the correct choice to use here? Do you think that maybe flexbox or grid would have been a better choice?

styles/style.css
```diff
@@ -120,18 +162,25 @@ p {
...
+  height: 20rem;
+  background-image: url(/images/header-bike.jpg);
   background-color: var(--grey-light);
```
MarcusZagorski: Not sure if background-color is needed. Maybe change this to background-size: cover, as this will ensure that the background takes up the complete space that your asking it to.

MarcusZagorski (commented):
Brilliant work Adniya, you have done an excellent job!

styles/style.css
AbdihamidAli: background-size: cover;
I tried this it still makes the image fitting need more adjustments 
we could use contain like this :point_down: 
background-size: contain;

styles/style.css
```diff
@@ -70,7 +72,7 @@ p {
   padding: .5rem 1em;
 }
 .navigation__item:hover{
-  background-color: var(--orange-light);
+  background-color: var(--orange-dark);
```
AbdihamidAli, line 81: this hover effect will make the navigation look like the buttons I think it should be only changing the color, not the background color

AbdihamidAli:
good work needs a few adjustments, especially the image and the navigation bar its hover effect, other than that great work :100:

styles/style.css
adniyaYousaf: Because it extends the full width of the available horizontal space of its parent element.

styles/style.css
adniyaYousaf: @MarcusZagorski. @AbdihamidAli  Thank you for bringing it to my attention. I've made the necessary corrections.

styles/style.css
adniyaYousaf: Yes, however, there are many websites where the links appear to have very similar styles.

PR #453, opened 2023-08-09
Wm5-bahadoryAhmadzay-Completed-Bike-for-Refugees-Week1

A-Bahadory:
Completed version Bike for Refugees. Much appreciated for any comments or feedback

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,179 +1,142 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -25,8 +25,9 @@
     </nav>
     <button class="btn-donate">Donate Now</button>
   </header>
+  
 
-
+<!-- main -->
```
MarcusZagorski, line 30: I think you can be a bit more specific with your comments. What actually is the main part? Is it the hero, the aside or the nav bar?

index.html
```diff
@@ -74,27 +75,20 @@ <h3 class="article__title">How can I help?</h3>
...
-
+</article>
 
       </section>
     </section>
 
-  </div>
 
 
     <aside class="sidebar">
```
MarcusZagorski, line 85: Attempt indenting your HTML better, as this will allow for more readability

MarcusZagorski (commented):
Great work Bahadory, you've laid everything out correctly. I would say to try and change the height of your background image to allow the hero image to look more similar to the image we were provided with

AbdihamidAli:
nice work, could you please add the hover effect to the buttons and the navigation bar?
And as Marcus pointed out to me that we need to change the buttons to anchors because those buttons would probably be linked to another page.

PR #455, opened 2023-08-09
WM5 | Marcus Zagorski | Module MTML-CSS | Week 1 

MarcusZagorski:
Here is my completed version of the Bike for refugees website. I have added the padding, images and all the other correct necessities pertaining to the website

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -163,7 +163,7 @@ p {
 .hero {
   background-image: url("/images/header-bike.jpg");
   background-color: var(--grey-light);
-  height: 55.2%;
+  height: 58%;
   background-size: auto;
   color: white;
   padding-left: 4.4%;
--- a/styles/style.css
+++ b/styles/style.css
@@ -76,7 +76,7 @@ p {
 
 .navigation__link:hover,
 .navigation__link:active {
-  color: #c05326;
+  color: var(--orange-dark);
 }
 
 .navigation__link {
@@ -105,7 +105,7 @@ p {
 
 /* Buttons */
 
-.donate-btn {
+.btn {
   background-color: #c05326;
   color: #e4e4e4;
   border: none;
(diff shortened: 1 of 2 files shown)


AbdihamidAli:
Great work Marcus.
If I'm not mistaken, wouldn't be better to have the entire image fit into the container?
Because sometimes the image may have some important details, if we add the image like we did now, the image will look like we cropped it.
Please let me know if this is a good idea 😊

MarcusZagorski:
> Great work Marcus. If I'm not mistaken, wouldn't be better to have the entire image fit into the container? Because sometimes the image may have some important details, if we add the image like we did now, the image will look like we cropped it. Please let me know if this is a good idea 😊

Could you elaborate on what you mean please Adbi? As far as I'm aware the image is contained inside of the Hero class, which I can edit the height, width and determine the side of the image inside of that container.

styles/style.css
```diff
@@ -120,8 +161,16 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("/images/header-bike.jpg");
   background-color: var(--grey-light);
+  height: 58%;
+  background-size: auto;
```
AbdihamidAli, line 161: I meant here by my comment earlier how about if we use this:
.hero{
background-size: contain;
}
when the "background-size is set to auto, the image still is cropped

AbdihamidAli: after doing some adjustments I found out we could do like this 
background-size: 100% 100%;
this property would stretch the image to fill the entire hero container.
the auto property the image would be cropped to fit the container by using the container width and height and ignoring the images width and height. If I'm not wrong :wrench:

MarcusZagorski: Thank you for noticing this Abdi! I'll fix this now

MarcusZagorski: After trying this out @AbdihamidAli I've come to notice that I didn't want to do what you said above as this will zoom the image out to fit inside of the container. But I appreciate your suggestion  🙏🏽

PR #457, opened 2023-08-10
Wm5/ElmiraZangeneh/HTML-CSS-SPRINT1-PREP

elmira-coding:


index.html
```diff
@@ -64,15 +64,18 @@ <h1>Bikes for Refugees</h1>
...
+              <button id="volunteer">Volunteer</button>
+            </div>
+         
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
-              <div class="article">
```
adniyaYousaf, line 75: Instead of Div element, it is recommended to use the Semantic HTML element  "Article".

adniyaYousaf (commented):
Congratulations on completing the task! You did a great job. Just use the Semantic HTML element.

AbdihamidAli:
Great work, as Marcus pointed out to me can you change the buttons as an anchor so it could be linked to another page? and also the Semantic HTML

PR #458, opened 2023-08-10
WM5/AISHA-ATHMAN-LALI/HTML-CSS-SPRINT1-PREPWORK

aishaathmanlali:
This is a website about Bikes for refugees. It was an interesting project. Please leave any feedback so that I can improve anything if need to.

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,7 +62,7 @@
             Providing donated bikes and accessories to refugees and asylum
             seekers in Scotland.
           </p>
-          <div class=" butt">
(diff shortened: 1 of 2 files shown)


AbdihamidAli:
Great work :100: , as Marcus pointed out to me can you change the buttons as an anchor so it could be linked to another page? and also the Semantic HTML

aishaathmanlali:
@AbdihamidAli Thank you for your suggestion. I think it is a good idea. I will try to do that.

PR #460, opened 2023-08-14
WM5-Sariat Adenike | Module HTML-CSS | Week-1

sariat0001:
First commit after fixing the page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,160 +20,164 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header">
+      <a href="#">
+        <img
+          src="images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees Scotland"
+          class="header__logo"
+        />
+      </a>
 
-        <a href="#">
-          <img
-            src="images/bikes-for-refugees_logo.jpg"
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -20,160 +20,164 @@
...
-            <li class="navigation__item">
-              <a class="navigation__link" href="#">News & Events</a>
-            </li>
-            <li class="navigation__item">
-              <a class="navigation__link" href="#">Contact Us</a>
-            </li>
-          </ul>
-        </div>
-
-        <button>Donate Now</button>
```
AbdihamidAli, line 50: Nice work, but I would recommend that if you could change the buttons to an anchor because we link them to another page. mostly we use anchors rather than a button if we only doing HTML code structure

AbdihamidAli:
Great work, I would say the bike image appears cropped from the bottom, I think because of the hero class you set its height to 50% and we need to change that to 58% so the image would be more shown in the container

sariat0001:
> Great work, I would say the bike image appears cropped from the bottom, I think because of the hero class you set its height to 50% and we need to change that to 58% so the image would be more shown in the container


Okay, Thanks for the review. I'll make the changes.

PR #461, opened 2023-08-15
WM5 | Karam Ali | Bikes For Refugees | Week 1

kayali436:
Completed version Bike for Refugees. 
Much appreciated for any comments or feedback

### DIFF
--- a/index.html
+++ b/index.html
@@ -25,7 +25,7 @@
     <header class="header">
       <a href="#">
         <img
-          src="/bikes-for-refugees/images/bikes-for-refugees_logo.jpg"
+          src="/images/bikes-for-refugees_logo.jpg"
           alt="Logo for Bikes for Refugees Scotland"
           class="header__logo"
         />
@@ -120,7 +120,7 @@
         <article class="article">
           <img
             class="article__thumbnail"
-            src="/bikes-for-refugees/images/spring-fundraisers_thumbnail.jpg"
+            src="/images/spring-fundraisers_thumbnail.jpg"
             alt="Spring fundraisers thumbnail"
           />
           <div class="article__content">
@@ -139,7 +139,7 @@
         <article class="article-aside">
           <img
             class="article__thumbnail"
-            src="/bikes-for-refugees/images/edinburgh-damascus_thumbnail.png"
+            src="/images/bikes-for-refugees_logo.jpg"
             alt="bike for refugees thumbnail"
           />
           <div class="article__content">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -19,161 +19,176 @@
...
-        </div>
-
-        <button>Donate Now</button>
-
-    </div>
+  <body>
+    <header class="header">
+      <a href="#">
+        <img
+          src="/bikes-for-refugees/images/bikes-for-refugees_logo.jpg"
```
SallyMcGrath: When you checked your work in the deploy preview, what showed up here?

Is this path working?

styles/style.css
```diff
@@ -92,11 +103,41 @@ p {
...
+  text-decoration: none;
+}
+
+.donate-btn:hover,
+.donate-btn:active {
+  background-color: white;
+  color: #c05326;
+}
 
+.volunteer-btn {
```
SallyMcGrath, line 125: Do you need to repeat all these styles again? What does class mean?

Can you make a class for all buttons, and then just set rules for the variations?

styles/style.css
```diff
@@ -120,8 +161,16 @@ p {
 /* hero */
 
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("/bikes-for-refugees/images/header-bike.jpg");
```
SallyMcGrath: Is this path correct? Where is this image relative to this stylesheet?

styles/style.css
```diff
@@ -120,8 +161,16 @@ p {
...
   background-color: var(--grey-light);
+  height: 58%;
+  background-size: auto;
+  color: white;
+  padding-left: 4.4%;
```
SallyMcGrath, line 169: This seems a bit like a kind of "eyeball" number or "magic number". What made you decide on 58% and 4.4%?

styles/style.css
```diff
@@ -221,17 +288,17 @@ p {
 
 /* Footer */
 
-.footer {
-  margin-top: 2rem;
+.footer_content {
+  margin-top: 10.8rem;
+  border-top: 2px solid var(--orange-light);
+  width: 76%;
```
SallyMcGrath, line 294: 76% of what? ;)

SallyMcGrath (commented):
Fab work, Karam!! ⭐ 

Thanks so much for your hard work on this. I've dropped a few comments and asked a few questions.  You might find this article interesting https://css-tricks.com/magic-numbers-in-css/

index.html
kayali436: Hi sorry I had a issue when pushing css file to my branch

PR #469, opened 2023-09-21
NW6 | Fathi-Kahin | HTML and CSS | Bikes for Refugees| week 1

fhkahin:
NW6- Fathi Updated Bikes for refugees

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,22 +14,22 @@
       rel="stylesheet"
       href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
     />
-    <link rel="stylesheet" href="styles/style.css" />
+    <link rel="stylesheet" href="/styles/style.css" />
 
     <title>Bikes for Refugees</title>
   </head>
 
   <body>
-    <div class="header">
+    <section class="header">
       <a href="#">
         <img
-          src="images/bikes-for-refugees_logo.jpg"
+          src="/images/bikes-for-refugees_logo.jpg"
           alt="Logo for Bikes for Refugees Scotland"
           class="header__logo"
         />
       </a>
 
-      <div>
+      <section>
         <ul class="navigation__list">
           <li class="navigation__item">
             <a class="navigation__link" href="#">Home</a>
@@ -44,33 +44,33 @@
             <a class="navigation__link" href="#">Contact Us</a>
           </li>
         </ul>
-      </div>
+      </section>
 
-      <button>Donate Now</button>
-    </div>
+      <button class="Don">Donate Now</button>
+    </section>
 
-    <div class="content">
-      <div class="main">
-        <div class="alert">
+    <section class="content">
+      <section class="main">
+        <section class="alert">
           You've helped us give <span class="text-highlight">72</span> bikes to
           refugees and asylum seekers so far. <strong>Thank you!</strong>
-        </div>
+        </section>
 
-        <div class="hero">
+        <section class="hero">
           <h1>Bikes for Refugees</h1>
           <p>
             Providing donated bikes and accessories to refugees and asylum
             seekers in Scotland.
           </p>
-          <button>Donate a bike today</button>
-          <button>Volunteer</button>
-        </div>
+          <button class="Don">Donate a bike today</button>
+          <button class="vol">Volunteer</button>
+        </section>
 
-        <div>
+        <section>
           <h2 class="heading-underline">Learn more</h2>
 
-          <div class="articles-container">
-            <div class="article">
+          <section id="art">
+            <section class="article" id="arts">
               <h3 class="article__title">Why do refugees need bikes?</h3>
               <p class="article__summary">
                 Many refugees are placed in housing in areas where there are few
@@ -83,9 +83,9 @@
                   Learn more
                 </a>
               </p>
-            </div>
+            </section>
 
-            <div class="article">
+            <section class="article" id="arts">
               <h3 class="article__title">How can I help?</h3>
               <p class="article__summary">
                 We need lots of bikes and bike accessories! If you have an old
@@ -98,21 +98,21 @@
                   Learn more
                 </a>
               </p>
-            </div>
-          </div>
-        </div>
-      </div>
+            </section>
+          </section>
+        </section>
+      </section>
 
-      <div class="sidebar">
+      <section class="sidebar">
         <h2 class="heading-underline">Upcoming events</h2>
 
-        <div class="article">
+        <section class="articles">
           <img
             class="article__thumbnail"
-            src="/images/spring-fundraisers_thumbnail.jpg"
-            alt="image of a tree branch with flowers in the spring bloom"
+            src="images/spring-fundraisers_thumbnail.jpg"
+            alt=""
           />
-          <div class="article__content">
+          <section class="article__content">
             <h3 class="article__title">
               <a class="article__title-link" href="#"
                 >Join us for our Spring fundraisers</a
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -14,166 +14,171 @@
...
     />
-    <link rel="stylesheet" href="styles/style.css" />
+    <link rel="stylesheet" href="/styles/style.css" />
 
     <title>Bikes for Refugees</title>
   </head>
 
   <body>
-    <div class="header">
+    <section class="header">
```
eliza-an, line 23: here you should have used a <header> tag!
Remember, the point of this exersice is SEMANTIC html. This means using html that has meaning. Section is very vague, while <header> tels the developer exactly what is is!

index.html
```diff
@@ -14,166 +14,171 @@
...
+      <a href="#">
+        <img
+          src="/images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees Scotland"
+          class="header__logo"
+        />
+      </a>
+
+      <section>
```
eliza-an, line 32: again, <seciton> is a very vague tag. You should use <nav> here. Nav stands for navigation! Here is a link https://www.w3schools.com/tags/tag_nav.asp

index.html
```diff
@@ -14,166 +14,171 @@
...
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">News & Events</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </section>
+
+      <button class="Don">Donate Now</button>
```
eliza-an, line 49: The class name "Don" is not very descriptive. Best practice is to use more descriptive words of what the element it. Next time, try to use "donate_Btn" or something so that it's more descriptive :)

index.html
```diff
@@ -14,166 +14,171 @@
...
+    </section>
+
+    <section class="content">
+      <section class="main">
```
eliza-an, line 53: you could use < main > here!

index.html
```diff
@@ -14,166 +14,171 @@
...
+                  <span class="arrow-right"></span>
+                  Learn more
+                </a>
+              </p>
+            </section>
+          </section>
+        </section>
+      </section>
+
+      <section class="sidebar">
```
eliza-an, line 106: here you can use a different tag instead of section

index.html
```diff
@@ -14,166 +14,171 @@
...
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </section>
+
+      <button class="Don">Donate Now</button>
+    </section>
+
+    <section class="content">
```
eliza-an, line 52: You could use a different more semantic tag here!

index.html
```diff
@@ -14,166 +14,171 @@
...
+          </p>
+          <button class="Don">Donate a bike today</button>
+          <button class="vol">Volunteer</button>
+        </section>
+
+        <section>
+          <h2 class="heading-underline">Learn more</h2>
+
+          <section id="art">
+            <section class="article" id="arts">
```
eliza-an, line 73: you could use  < article > here! Try researching some more semantic HTML

index.html
```diff
@@ -14,166 +14,171 @@
...
+              </p>
+              <p class="article__read-more">
+                <a href="#">
+                  <span class="arrow-right"></span>
+                  Learn more
+                </a>
+              </p>
+            </section>
+
+            <section class="article" id="arts">
```
eliza-an, line 88: You could use a different more semantic hTML tag here! like <article >

index.html
```diff
@@ -14,166 +14,171 @@
...
+            </section>
+          </section>
+        </section>
+      </section>
+
+      <section class="sidebar">
+        <h2 class="heading-underline">Upcoming events</h2>
 
-        <a href="#">
+        <section class="articles">
```
eliza-an, line 109: You could use a different tag here!

index.html
```diff
@@ -14,166 +14,171 @@
...
-            You've helped us give <span class="text-highlight">72</span> bikes
-            to refugees and asylum seekers so far. <strong>Thank you!</strong>
-          </div>
-
-          <div class="hero">
-            <h1>Bikes for Refugees</h1>
-            <p>
-              Providing donated bikes and accessories to refugees and asylum
-              seekers in Scotland.
+        <section class="articles">
```
eliza-an, line 128: you could use a different tag here

index.html
```diff
@@ -14,166 +14,171 @@
...
+          <section class="article__content">
+            <h3 class="article__title">
+              <a class="article__title-link" href="#"
+                >Bike dropoff event in Edinburgh on May 11</a
+              >
+            </h3>
+          </section>
+        </section>
+
+        <section class="articles">
```
eliza-an, line 143: yuou could use a different tag here!

index.html
```diff
@@ -14,166 +14,171 @@
...
+          <img
+            class="article__thumbnail"
+            src="images/edinburgh-damascus_thumbnail.png"
+            alt=""
+          />
+          <section class="article__content">
```
eliza-an, line 149: You could use a different tag here!

index.html
```diff
@@ -14,166 +14,171 @@
...
-            <p>
-              Providing donated bikes and accessories to refugees and asylum
-              seekers in Scotland.
+        <section class="articles">
+          <img
+            class="article__thumbnail"
+            src="images/bikes-for-refugees_logo.jpg"
+            alt=""
+          />
+          <section class="article__content">
```
eliza-an, line 134: You could use a different tag here!

index.html
```diff
@@ -14,166 +14,171 @@
...
           <img
-            src="images/bikes-for-refugees_logo.jpg"
-            alt="Logo for Bikes for Refugees Scotland"
-            class="header__logo"
+            class="article__thumbnail"
+            src="images/spring-fundraisers_thumbnail.jpg"
+            alt=""
           />
-        </a>
+          <section class="article__content">
```
eliza-an, line 115: You could use a different tag here!

styles/style.css
```diff
@@ -23,8 +23,36 @@ button {
...
 }
-
-h1,h2,h3,h4,h5,h6 {
+.Don {
+  color: white;
+  background: #c05326;
+  padding: 7px;
+  margin-left: 17px;
+  margin-bottom: 100px;
+  margin-top: 30px;
```
eliza-an, line 32: Look at your website and the picture
Do you see how the "Donate Now" button on the picture is inline with the navigation bar? 
How can you fix this? 
Also, in the picture, the donate now button has rounded corners whilst yours is very sharp, try using the ( border-radius:) to change this!
Here is what is should look like! : 

image

index.html
```diff
@@ -14,166 +14,171 @@
...
+          refugees and asylum seekers so far. <strong>Thank you!</strong>
+        </section>
+
+        <section class="hero">
+          <h1>Bikes for Refugees</h1>
+          <p>
+            Providing donated bikes and accessories to refugees and asylum
+            seekers in Scotland.
+          </p>
+          <button class="Don">Donate a bike today</button>
```
eliza-an, line 65: You gave two buttons the same class! This is a problem because if I want to change something about only one of the buttons, i cannot! 
The space between this button and the rest of the test is too large! because you added a margin to that css! You have to change this class and add new css to this!
 
image

index.html
```diff
@@ -14,166 +14,171 @@
...
-              </p>
-            </div>
-          </div>
+          </section>
+        </section>
+
+        <a
+          href="https://www.facebook.com/BikesforRefugeesScotland/events/?ref=page_internal"
+          target="_blank"
+          class="facebook-link"
```
eliza-an, line 163: This link is on the right on your page, but it is on the left in the picture! try adding display:flex
and then play around with justify-content:  to put it on the left

eliza-an (changes requested):
This was an excellent first attempt. The only issue i noticed was that you changed every <div> to <section> instead of using more semantic HTML. Next time, use semantic html like <main> and <aside>. Semantic HTML is HTML that has meaning. <section can be used sometimes but its very vague! Here is a good resource for this:  https://www.w3schools.com/html/html5_semantic_elements.asp

PR #470, opened 2023-09-22
NW6/OrlandoMorales/HTML&CSS/BikesForRefugees/Week1

OrlandoMoralesKuan:
I did some additions and substraction in HTML and CSS to try to make the website looks like the task


bunday (commented):
This is a very good work. Just a little comment. When I hover on the links on the header, the background should be white not block

index.html
```diff
@@ -64,14 +65,14 @@ <h1>Bikes for Refugees</h1>
...
-            <button>Volunteer</button>
+            <button class="donate">Donate a bike today</button>
+            <button class="volunteer">Volunteer</button>
           </div>
 
           <div>
             <h2 class="heading-underline">Learn more</h2>
 
-            <div>
+            <div class="articleArea">
```
SpectacularLandscape, line 75: For many of these `<div>` elements that have a class similar to `article`, have a look through these [semantic HTML elements](https://developer.mozilla.org/en-US/docs/Glossary/semantics#semantic_elements) and see if there is one that fits. It is very similar to the name already in these classes!

index.html
```diff
@@ -87,7 +88,7 @@ <h3 class="article__title">Why do refugees need bikes?</h3>
                 </p>
               </div>
 
-              <div class="article">
+              <div class="article1">
```
SpectacularLandscape, line 91: I can see `<div>` elements on the page with very similar (but not quite identical) classes, for example, `articleArea`, `article`, `articles`. Classes are used where we want to style multiple elements in a similar way. IDs are used when there is only one element that should be styled. Have a think about whether these elements with "article"-like classes should all have the same class or not

PR #471, opened 2023-09-22
NW6 | Areeb Sattar | HTML-CSS | Week-1

areebsattar:


SpectacularLandscape:
Great work Areeb! You've successfully completed all the tasks and even the stretch task, and it's clear you understand how to use semantic HTML and CSS styles. I especially like the use of `<nav>` in your `<header>` 🥇

My only suggestion is to take the semantic HTML a little bit further. The "Upcoming Events" section suits a `<section>` containing `<article>`s as well. And like headers, footers have their own semantic element that is (unsurprisingly) `<footer>`

Keep up the great work, I look forward to seeing what comes next!

PR #474, opened 2023-09-22
NW6 |Sabella-Fisseha | HTML-CSS | bike-for-refugee | week1

Sabella-8:
NW6/Sabella Goytom Fisseha


SpectacularLandscape:
This looks great Sabella! You've completed all of the tasks and the stretch goal as well 🌟 

I've left a few comments on the code changes to help your HTML be even more semantic 😄

index.html
```diff
@@ -20,160 +20,165 @@
   </head>
 
   <body>
-    <div class="header">
+    <section class="header">
```
SpectacularLandscape, line 23: There's a dedicated HTML element called [`<header>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) that is great to use here. This would help to reduce the number of CSS classes you need to use as well, as you could directly reference your `header` element

index.html
```diff
@@ -20,160 +20,165 @@
...
+      <a href="#">
+        <img
+          src="images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees Scotland"
+          class="header__logo"
+        />
+      </a>
+
+      <section>
```
SpectacularLandscape, line 32: Like the `<header>`, we also have a dedicated element for navigation called [`<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav)

index.html
```diff
@@ -20,160 +20,165 @@
...
+            seekers in Scotland.
+          </p>
+          <button class="dovo">Donate a bike today</button>
+          <button class="vol">Volunteer</button>
+        </section>
+
+        <section>
+          <h2 class="heading-underline">Learn more</h2>
+
+          <section id="art">
```
SpectacularLandscape, line 72: This would be a great place to use the [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) element

index.html
```diff
@@ -20,160 +20,165 @@
...
-          >
-            <span class="fa fa-facebook-square"></span>
-            All Facebook events
-          </a>
-        </div>
-    </div>
-
-    <div class="footer">
-      <div class="footer__content">
+    <section class="footer">
```
SpectacularLandscape, line 171: We have [`<footer>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer) available to use here

index.html
```diff
@@ -20,160 +20,165 @@
...
+            <a class="navigation__link" href="#">Contact Us</a>
+          </li>
+        </ul>
+      </section>
+
+      <button class="dovo">Donate Now</button>
+    </section>
+
+    <section class="content">
+      <section class="main">
```
SpectacularLandscape, line 53: As a rule of thumb, try to keep `<section>` elements only for blocking out sections of your web page. The [MDN Web Docs page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) suggest only using a section where you can use a heading element (`<h1>`, `<h2>`, etc.), with a few exceptions for grouping together content.

For example, on this page you could use a `<section>` for "Learn more" and "Upcoming events". And you might also use a `<section>` to separate out the 'hero' area of the page - even if that doesn't have a dedicated heading.

Elsewhere, you can use `<div>` sections to add styles to the page

PR #476, opened 2023-09-22
NW6 | Hadika Malik | HTML-CSS | Bikes-for-refugees | Week 1 

HadikaMalik:


index.html
```diff
@@ -20,160 +20,165 @@
...
+        <img
+          src="images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees Scotland"
+          class="header__logo"
+        />
+      </a>
+
+      <nav>
+        <ul class="navigation__list">
+          <li class="navigation__item">
```
bunday, line 34: on hover of these links, instead of adding an orange background with black colour for the text. You should leave the background as white and then change the link text colour to orange
Screenshot 2023-09-26 at 09 22 05

bunday (changes requested):
Great work, kindly adjust your PR title to reflect the current format discussed

PR #477, opened 2023-09-23
NW6|Bakhat Begum|HTML_CSS|Bikes-for-refugees| Week 1

BakhatBegum:
I replaced the div's with semantic HTML tags.
 I replace broken image links.
I added a style button.
I added Flexbox and hovered over the article class.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,7 +24,7 @@
       <a href="#">
         <img
           src="images/bikes-for-refugees_logo.jpg"
-          alt="Logo for Bikes for Refugees Scotland"
+          alt=" promote bikes for refugees scotland"
           class="header__logo"
         />
       </a>
@@ -66,11 +66,11 @@
           <button class="btn">Volunteer</button>
         </figcaption>
 
-        <article>
+        <section>
           <h2 class="heading-underline">Learn more</h2>
 
-          <article class="learn-more-articles">
-            <section class="block">
+          <section class="learn-more-articles">
+            <article class="block">
               <h3 class="article__title">Why do refugees need bikes?</h3>
               <p class="article__summary">
                 Many refugees are placed in housing in areas where there are few
@@ -83,9 +83,9 @@
                   Learn more
                 </a>
               </p>
-            </section>
+            </article>
 
-            <section class="block">
+            <article class="block">
               <h3 class="article__title">How can I help?</h3>
               <p class="article__summary">
                 We need lots of bikes and bike accessories! If you have an old
@@ -98,19 +98,19 @@
                   Learn more
                 </a>
               </p>
-            </section>
-          </article>
-        </article>
+            </article>
+          </section>
+        </section>
       </main>
 
-      <article class="sidebar">
+      <section class="sidebar">
         <h2 class="heading-underline">Upcoming events</h2>
 
         <aside class="article">
           <img
             class="article__thumbnail"
             src="images/spring-fundraisers_thumbnail.jpg"
-            alt=""
+            alt="promote the fundraisers"
           />
 
           <figcaption class="article__content">
@@ -130,7 +130,7 @@
           <img
             class="article__thumbnail"
             src="images/bikes-for-refugees_logo.jpg"
-            alt="bikes image"
+            alt="promote drop off event"
           />
           <figcaption class="article__content">
             <h3 class="article__title">
@@ -145,7 +145,7 @@
           <img
             class="article__thumbnail"
             src="images/edinburgh-damascus_thumbnail.png"
-            alt="image for Help us cycle 4,797km"
+            alt="promote for Help us distance 4,797km"
           />
           <figcaption class="article__content">
             <h3 class="article__title">
@@ -166,7 +166,7 @@
           <span class="fa fa-facebook-square"></span>
           All Facebook events
         </a>
-      </article>
+      </section>
     </content>
 
     <footer class="footer">
--- a/index.html
+++ b/index.html
@@ -20,160 +20,166 @@
   </head>
 
   <body>
-    <div class="header">
+    <header class="header">
+      <a href="#">
+        <img
+          src="images/bikes-for-refugees_logo.jpg"
+          alt="Logo for Bikes for Refugees Scotland"
+          class="header__logo"
+        />
+      </a>
 
-        <a href="#">
+      <nav>
+        <ul class="navigation__list">
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">Home</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">About Us</a>
+          </li>
+          <li class="navigation__item">
+            <a class="navigation__link" href="#">News & Events</a>
+          </li>
+          <li class="navigation__item">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -20,160 +20,166 @@
...
-        </div>
-    </div>
-
-    <div class="footer">
-      <div class="footer__content">
+        <aside class="article">
+          <img
+            class="article__thumbnail"
+            src="images/bikes-for-refugees_logo.jpg"
+            alt="bikes image"
```
eliza-an, line 133: You don't need to say "image" in the alt tag. The alt tag is made specifically to describe images, so adding the word image is unnecessary. It's important to keep the information as concise but descriptive as possible here!

index.html
```diff
@@ -20,160 +20,166 @@
...
+              >
+            </h3>
+          </figcaption>
+        </aside>
+
+        <aside class="article">
+          <img
+            class="article__thumbnail"
+            src="images/edinburgh-damascus_thumbnail.png"
+            alt="image for Help us cycle 4,797km"
```
eliza-an, line 148: same "image" comment as up there ^

index.html
```diff
@@ -20,160 +20,166 @@
...
 
-        <a href="#">
+        <aside class="article">
           <img
-            src="images/bikes-for-refugees_logo.jpg"
-            alt="Logo for Bikes for Refugees Scotland"
-            class="header__logo"
+            class="article__thumbnail"
+            src="images/spring-fundraisers_thumbnail.jpg"
+            alt=""
```
eliza-an, line 113: There is no alt tag here!
Also there are some comments in the css folder about this image styling. Pls check the css file!

index.html
```diff
@@ -20,160 +20,166 @@
...
-            to refugees and asylum seekers so far. <strong>Thank you!</strong>
-          </div>
-
-          <div class="hero">
-            <h1>Bikes for Refugees</h1>
-            <p>
-              Providing donated bikes and accessories to refugees and asylum
-              seekers in Scotland.
+          <figcaption class="article__content">
+            <h3 class="article__title">
```
eliza-an, line 117: have a look at the image provided to you. Your title looks a bit squashed to the top. How would you fix that? Maybe with some padding?

index.html
eliza-an: There is no alt tag here!
Also, have a look at your page and the page given. Does yours look the same? Try playing around with the padding of this element in the CSS!

index.html
eliza-an: You don't need to say "image" in the alt tag. The alt tag is made specifically to describe images, so adding the word image is unnecessary. It's important to keep the information as concise but descriptive as possible here!

index.html
eliza-an: same "image" comment as up there ^

styles/style.css
```diff
@@ -92,12 +104,24 @@ p {
   border-left-color: var(--orange-dark);
 }
 
-
 /* Buttons */
+button {
+  background-color: #c05326;
+  color: white;
+  border-radius: 5px;
+}
```
eliza-an, line 111: in here, add " border: none "  here to remove the black outline on your buttons. Do you see how yours has one and the sample image doesnt?

eliza-an, line 111: I would also add a little more padding here to make the buttons a little larger as yours are quite thin!

index.html
```diff
@@ -20,160 +20,166 @@
...
+              We need you to help us cycle 4797km, the distance of Edinburgh to
+              Damascus.
+            </p>
+          </figcaption>
+        </aside>
+
+        <a
+          href="https://www.facebook.com/BikesforRefugeesScotland/events/?ref=page_internal"
+          target="_blank"
+          class="facebook-link"
```
eliza-an, line 164: for the class "facebook-link" 
Have a look at the sample image vs your website. Yours is to the left whilst theirs is to the right. How could you change that? Try adding display:flex  to this class in css and play around with the grid inspector in dev tools to figure out what else you have to add!

styles/style.css
```diff
@@ -186,6 +230,15 @@ p {
...
 }
 
+.block {
+  display: flex;
+  flex-direction: column;
+  border-style: solid;
+  width: 45%;
+  border-color: var(--grey-light);
+  align-content: center;
+}
```
eliza-an, line 245: I would add some padding here becuase your text is a little squashed inside your box. add some spacing to it! Have a look at the design and see what would be best!

styles/style.css
```diff
@@ -158,17 +191,28 @@ p {
...
+  border-color: var(--grey-light);
+  border-style: solid;
 }
 
 .article__title {
   margin-bottom: 0.5rem;
   font-size: 1rem;
-  font-weight: 700;
+  font-weight: 600;
+  padding-left: 20px;
```
eliza-an: If you have a look at the sample image, do you see how their lines align at the sides? You dont really want your title to stick out the way yours does! 
Their image: 
image

Yours: 
image


How can you fix this? Have a look at your padding-left property here!

styles/style.css
```diff
@@ -108,7 +132,9 @@ p {
 /* Alert */
 
 .alert {
+  border-style: solid;
```
eliza-an: The sample image does not have a border here so that's not needed :)

styles/style.css
```diff
@@ -59,15 +64,19 @@ p {
 }
 
 /* Navigation */
+nav {
+  padding-right: 30rem;
```
eliza-an: Do you see how your navbar skews to the left? That's because of this padding-right there! Try removing it and see if it looks more like the image

styles/style.css
```diff
@@ -92,12 +104,24 @@ p {
...
-  /* INSERT BUTTON STYLES HERE */
+/* INSERT BUTTON STYLES HERE */
 
+button:hover {
+  background-color: gray;
+}
 
+.btn {
+  background-color: white;
+  color: orange;
```
eliza-an: For the sake of the page looking more synchronous, you always want to use similar colors. This orange doesn't match the one we used for the other button! Play around with the colors and the font-weight of this!!

styles/style.css
```diff
@@ -120,21 +146,27 @@ p {
...
 .hero {
-  background-image: url("header-bike.jpg");
+  background-image: url("../images/header-bike.jpg");
+  width: 95%;
+  height: 350px;
   background-color: var(--grey-light);
+  padding-left: 30px;
 }
 
 .hero h1 {
```
eliza-an, line 154: Play around with the styling: padding and margins here. This doesnt look exactly like the picture!

eliza-an (changes requested):
A good attempt, but for next time try to pay more attention to the details! When you are a developer, that's your main job, to recreate the image they gave you exactly! You want to notice all of the small details. A good first attempt overall! You did well with the semantic HTML, adding the images and using the hover effect,  but the rest of the criteria needs a little more work! :)

BakhatBegum:
I made changes in all the syntax that you have mentioned.

index.html
BakhatBegum: Hello Eliza, I committed all those mistakes that you mentioned.

PR #483, opened 2023-10-07
nw6 | nazanin-saedi| BikeForRefugees | [TECH ED] Bikes for Refugees | week1

nazaninsaedi:


index.html
```diff
@@ -64,14 +64,13 @@ <h1>Bikes for Refugees</h1>
               Providing donated bikes and accessories to refugees and asylum
               seekers in Scotland.
             </p>
-            <button>Donate a bike today</button>
-            <button>Volunteer</button>
+            <button class="donated">Donate a bike today</button>
```
Ara225, line 67: Descriptive class names :)  You need to style these a bit diffrently though to look like the image but I am sure you are aware of that. Were also need background images

index.html
```diff
@@ -108,10 +109,10 @@ <h3 class="article__title">How can I help?</h3>
         <div class="sidebar">
           <h2 class="heading-underline">Upcoming events</h2>
 
-          <div class="article">
+          <div class="articles">
             <img
               class="article__thumbnail"
-              src="spring-fundraisers_thumbnail.jpg"
+              src="/images/spring-fundraisers_thumbnail.jpg"
```
Ara225, line 115: Good fix

style.css
```diff
@@ -0,0 +1,295 @@
+/**
+ *
+ * Module: HTML/CSS
+ * Lesson: 1
+ * 
+ */
+
+ :root {
+  --grey-dark: #292b2c;
+  --grey-light: #e4e4e4;
```
Ara225, line 10: Good that you're using CSS vars :)

style.css
```diff
@@ -0,0 +1,295 @@
...
+ font-size: 0.85rem;
+ line-height: 1.3;
+ padding: 1.5rem 1rem;
+ margin-bottom: 2rem;
+}
+
+/* hero */
+
+.hero {
+ background-image: url("../images/header-bike.jpg");
```
Ara225, line 124: I'm afraid this link is broken, it will work in the other CSS file

style.css
```diff
@@ -0,0 +1,295 @@
...
+
+.content {
+ display: flex;
+ flex-direction: column;
+}
+
+/* Alert */
+
+.alert {
+ background-color: var(--orange-light);
```
Ara225, line 112: All the styles here look really good

style.css
```diff
@@ -0,0 +1,295 @@
+/**
+ *
```
Ara225, line 2: This is the wrong CSS file I'm afraid - the one in syles is the one you want this is why the styles aren't showing up. I think the work you have done is very good though, it's very close
https://github.com/CodeYourFuture/bikes-for-refugees/blob/main/styles/style.css

Ara225 (changes requested):
Hello!
Congrats on your first pull request! Seems liek you've done most the work, you just need to move style.css into the styles folder and you'll be almost there :) Thanks

MaryamAlattal (commented):
Good work 👍 suggestions: 
Navbar should be properly styled (Position/spacing/Color text/hover effects)

MaryamAlattal (commented):
Well done to use background-image in class hero👍 Recommendations:

Check the correct path (../images/etc); With the same class hero, the style of the buttons needs to be changed

PR #484, opened 2023-10-07
NW6/Ayman-matoo/Html Css /Bikes for refugees /week 1

Ayman-matoo:


index.html
```diff
@@ -21,158 +21,174 @@
 
   <body>
     <div class="header">
```
Honestocean: This div isn't semantic, it's a generic html tag. Mayble look at the class string as a clue to make it semantic!

index.html
```diff
@@ -21,158 +21,174 @@
...
-            </p>
-            <button>Donate a bike today</button>
-            <button>Volunteer</button>
+        <div class="hero">
+          <h1>Bikes for Refugees</h1>
+          <p>
+            Providing donated bikes and accessories to refugees and asylum
+            seekers in Scotland.
+          </p>
+          <div class="button-div">
```
Honestocean, line 67: You don't need to add this div in. As mentioned the divs are not semantic, whereas the button tag is.

index.html
Ayman-matoo: okay, thank

PR #531, opened 2025-09-08
London/September 2025/alexandru pocovnicu/bikes-for-refugees

alexandru-pocovnicu:
Changed HTML elements to semantic HTML and added style to it


alexandru-pocovnicu:
I haven't managed to widen the .hero background image to make it look similar to the one in the example, I tried using vh and % but to no avail.

